PR #9, opened 2015-09-17
Thank Robert Jördens for the PyPi package

marcua:


thisisdhaas:
I think the `thank you` section title might be better as something like `acknowledgments` or the like, but we can change it when we actually write up the readme.

LGTM.

PR #26, opened 2015-09-24
Add the admin back in.

thisisdhaas:
(So people can create certifications and other necessary stuff that isn't in the API yet.)


marcua:
lgtm

PR #34, opened 2015-09-28
Add API reference and update docs from website copy

jrbotros:


docs/source/api.rst
```diff
@@ -1,17 +1,105 @@
...
-    :members:
-
-.. task_lifecycle:
-Task Lifecycle
---------------
-.. automodule:: orchestra.utils.task_lifecycle
-    :members:
+Client API
+----------
+
```
marcua, line 19: add a note that they have to sign the request using http signatures (http://tools.ietf.org/html/draft-cavage-http-signatures-03)

See an example in `_make_api_request` of https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/orchestra_api.py#L29

docs/source/concepts.rst
```diff
@@ -1,2 +1,78 @@
 Key Concepts
 ============
+
+Let's first recap our example reporting workflow:
```
marcua, line 4: include the diagram as @thisisdhaas does in his PR

docs/source/concepts.rst
```diff
@@ -1,2 +1,78 @@
...
+- A senior photographer reviews the photos and selects the best ones.
+- The selected photos are resized and recolored for display across different media.
+- Finally, a copy editor adds headlines and photo captions to complete the story.
+
+We'll now walk you through major Orchestra concepts based on the example above.
+
+
+Workflows
+---------
+- The entire process above is called a workflow, comprised of five component steps.
```
marcua: we should preserve all of the formatting like bolding/italics from the website.

docs/source/concepts.rst
```diff
@@ -1,2 +1,78 @@
...
+
+Life of a Task
+--------------
+Below are two images of the Orchestra dashboard, the launching point for expert workers.
+Click to see how tasks move differently across the dashboard for core workers and reviewers.
+
+
+Core Expert
+~~~~~~~~~~~
+A core expert performs initial task work which will later be reviewed. The below diagram shows a task's movement through the core worker's dashboard.
```
marcua: diagram below

docs/source/concepts.rst
```diff
@@ -1,2 +1,78 @@
...
+
+.. image:: ../static/img/core_lifecycle.jpg
+
+
+Reviewer
+~~~~~~~~
+A reviewer evaluates the core expert's work and provides feedback. The below diagram shows a task's movement through a reviewer's dashboard.
```
marcua: diagram below

docs/source/core.rst
```diff
@@ -0,0 +1,26 @@
+Core Reference
+=============
```
marcua, line 2: why not just all of `orchestra` and `beanstalk_dispatch`?

docs/source/concepts.rst
jrbotros, line 4: If going in order, they'll have just come from the example use case section—I find it pretty redundant to duplicate the image here. I can put some text linking back to that section at the top?

docs/source/concepts.rst
jrbotros: :white_check_mark:

docs/source/concepts.rst
jrbotros: :white_check_mark:

docs/source/core.rst
jrbotros, line 2: I've only included the modules I took the time to carefully docstring—don't think we should put the others in until they're cleaned up. Maybe we should add a "core reference in progress" qualifier somewhere?

docs/source/concepts.rst
jrbotros: :white_check_mark:

docs/source/api.rst
jrbotros, line 19: :white_check_mark:

docs/source/concepts.rst
marcua, line 4: there is no such thing as redundancy in documentation, and we have no guarantee someone will read it in the order we suggested.  totally reasonable for someone to read the key concepts section first, and we should have an example there waiting for them.  Otherwise, aren't the bullet points just as redundant as the image?:P

docs/source/core.rst
marcua, line 2: i think so---it's otherwise ridiculously sparse :)

docs/source/concepts.rst
jrbotros, line 4: :white_check_mark:

docs/source/core.rst
jrbotros, line 2: :white_check_mark:

marcua:
LGTM!

PR #43, opened 2015-10-07
Add signal registration for task submit and save

jrbotros:


orchestra/static/orchestra/task/js/controllers.js
```diff
@@ -116,28 +116,33 @@
...
         })
         .finally(function() {
-          vm.saving = false;
           if (vm.saveError) {
             // Retry save with exp backoff
             vm.autoSaveTimeout *= 2;
             vm.scheduleAutoSave();
           }
+          orchestraService.signals.save.finally();
+          vm.saving = false;
```
marcua: ?????

orchestra/static/orchestra/task/js/controllers.js
```diff
@@ -146,10 +151,13 @@
           // will be saved by submission
           vm.cancelAutoSave();
           $location.path('/');
```
marcua, line 154: call functions before changing location, but after canceling autosave?

orchestra/static/orchestra/task/js/controllers.js
```diff
@@ -116,28 +116,33 @@
...
+          orchestraService.signals.save.error();
         })
         .finally(function() {
-          vm.saving = false;
           if (vm.saveError) {
             // Retry save with exp backoff
             vm.autoSaveTimeout *= 2;
             vm.scheduleAutoSave();
           }
+          orchestraService.signals.save.finally();
```
marcua: sanity check that this should go here and not earlier in this block

orchestra/static/orchestra/task/js/services.js
```diff
@@ -47,9 +47,26 @@ serviceModule.factory('orchestraService', function() {
         }
     }
 
+    var signals = {
+        // Functions to override in specific task controllers.
+        'save': {
```
marcua: high-level: we don't want callers to override this, because we might want two callers to subscribe to signals

the way this can happen is:
- we don't expose `signals`---that stays internal
- we expose a function `registerSignal(signalType, state, callback)` that does `signals[signalType][signal].append(callback)` (Python syntax, I know :))
- we have an internal function called `fireSignal(signalType, state)` that loops through `signals[signalType][signal]` and calls the callbacks.
- replace all of the calls above, e.g., `orchestraService.signals.save.finally();` -> `fireSignal(signalType, state);`
- i'm not even sure we need to separate `signalType` and `state`: you could imagine having a flatter structure, and we can name signals things like `submit.before`

orchestra/static/orchestra/task/js/controllers.js
jrbotros: Originally wanted the `saving` flag to be reset at the very end, but it probably makes sense to change it before the next save is scheduled if there's an error!

orchestra/static/orchestra/task/js/controllers.js
jrbotros, line 154: Oops, that definitely makes sense 😳

orchestra/static/orchestra/task/js/controllers.js
jrbotros: See comment below :white_check_mark:

orchestra/static/orchestra/task/js/services.js
jrbotros: Sounds good to me—I like the `registerSignal('submit.before', function() {...})` style, so I'll take a stab at that now!

orchestra/static/orchestra/common/js/lib/sparkline-chart.js
```diff
@@ -123,4 +123,4 @@ var Script = function () {
 
     });
 
-}();
\ No newline at end of file
+}();
```
jrbotros: Oops, my Sublime wanted in on the action. Will take this out before merge!

marcua:
lgtm

PR #44, opened 2015-10-08
Integrate quilljs component for custom task views

jrbotros:
Adds the [Quill editor](https://github.com/quilljs/quill/) as an optional directive to be included in custom task views.


orchestra/static/orchestra/common/components/modules.js
```diff
@@ -0,0 +1,11 @@
+(function () {
```
marcua, line 1: what's the benefit of the components module?

marcua:
Quick sniff test lgtm.  One question on here.

It might be nice to include a quick example html snippet of how to include a quill editor in a task, and an update to the README to explain its dependence on things like S3, how to configure it, etc.

# 

@jrbotros
This makes sense. I included `components-example.html`—take a look and let me know what you think!

orchestra/static/orchestra/common/components/quill/js/directives.js
```diff
@@ -0,0 +1,214 @@
...
+          var contents = scope.editor.getContents()
+          contents.ops = contents.ops.slice(0, opIndex)
+          return contents.length() + caretPosition.offset;
+         }
+
+        // Post image data to an API endpoint that returns an image URL, then
+        // adding it to the editor (replacing the given character range)
+        function uploadImage(file, range, e) {
+          var uploadAPIEndpoint = '/orchestra/api/interface/upload_image/'
+          var maxUploadSizeMB = 20;
```
marcua: large, no ?:)

orchestra/static/orchestra/common/components/modules.js
jrbotros, line 1: The idea was to nicely collect all the reusable task view components (e.g., rich text editor, checklists) into a single directory. I nested it inside `common` because I can imagine there would be other unrelated common directives we'd want to include.

Maybe `components` should live inside the `task` static files instead, since they'll probably only be used in custom task views? Though you could imagine using a rich text editor in the admin or something...

orchestra/static/orchestra/common/components/quill/js/directives.js
jrbotros: Yeah, a bit, and also application-dependent—maybe we should just pass this into the directive scope?

orchestra/static/orchestra/common/components/modules.js
marcua, line 1: Your idea sounds good to me.  Let's not move it into `task` until we get the sense that makes sense :).

orchestra/static/orchestra/common/components/quill/js/directives.js
marcua: that's fine, but as a first pass, make it 5 MB and we can handle it later?

orchestra/static/orchestra/common/components/modules.js
jrbotros, line 1: :white_check_mark:

orchestra/static/orchestra/common/components/quill/js/directives.js
jrbotros: I went ahead and added it as an optional directive attribute with 5 MB as a fallback, let me know if that works!

marcua:
Three comments, but you can fix/squash/merge w/o another review assuming the fixes are simple/obvious :)

PR #50, opened 2015-10-13
Task assignment iteration latency reporting

marcua:
This PR exposes a `orchestra.analytics.latency`, a module with methods that help us study task assignment iteration calendar time and work latencies.

It doesn't include any sort of aggregation/command-line tooling---that might come in a followup PR.

### DIFF
--- a/orchestra/analytics/latency.py
+++ b/orchestra/analytics/latency.py
@@ -1,0 +1,125 @@
+import itertools
+from datetime import timedelta
+from dateutil.parser import parse
+from operator import attrgetter
+from orchestra.models import Task
+from orchestra.project_api.api import get_project_information
+from orchestra.workflow import Step
+from pandas import DataFrame
+
+
+def work_time_df(projects):
+    """
+    Return projects' task assignment iteration timing information.
+
+    The returned dataframe's schema is:
+        project_id: the project's ID
+        task_id: the task's ID
+        assignment_level: a 1-indexed assignment counter (1 is entry-level, 2
+            is first review, etc.)
+        worker: the username of the worker on that assignment
+        iteration: a work iteration for the assignment:
+            - Iteration 0 represents the period from previous submission to
+            reviewer assignment:
+                - If this is the first assignment, 'previous submission'
+                  means 'task creation'.
+                - For iteration 0, work time will always be 0
+                - If this task is auto-assigned, iteration 0's calendar time
+                  will be close to 0.
+            - Iterations N > 0 represent the period between a previous worker's
+                input (submit/accept/reject) and the worker's submission.
+        start_datetime: the time the iteration started.
+        calendar_time: the wallclock time it took for the worker to complete
+            that iteration.
+        work_time: the amount of work time the worker reported dedicating to
+            that iteration (0 for iteration 0).
+
+    Args:
+        projects (django.db.models.query.QuerySet):
+            An iterable of projects to include in the report.
+
+    Returns:
+        df (pandas.DataFrame):
+            A DataFrame with timing information.
+    """
+    # TODO(marcua): `get_project_information` per project means we do random
+    # access all over the database.  Ideally, we'd have a sequential scan
+    # version of this.
+    time_data = list(itertools.chain.from_iterable(
+        project_time_row_generator(
+            get_project_information(project.id))
+        for project in projects))
+    df = DataFrame(time_data)
+    return df
+
+
+class Iteration(object):
+    """ Represents a task assignment iteration's data for reporting. """
+    def __init__(self, project, task, assignment, iteration, worker,
+                 work_time, end_datetime):
+        self.project = project
+        self.task = task
+        self.assignment = assignment
+        self.iteration = iteration
+        self.work_time = work_time
+        self.end_datetime = end_datetime
+        self.worker = worker
+
+    def to_dict(self, start_datetime):
+        return {
+            'project_id': self.project['id'],
+            'task_id': self.task['id'],
+            'assignment_level': self.assignment,
+            'iteration': self.iteration,
+            'start_datetime': start_datetime,
+            'work_time': self.work_time,
+            'calendar_time': self.end_datetime - start_datetime,
+            'worker': self.worker
+        }
+
+
+def project_time_row_generator(project_information):
+    """
+    Given a project, yields all task assignment iterations, reporting
+    calendar and work times.
+    """
+    # lol at this variable name
+    human_slugs = {
+        step['slug'] for step in project_information['steps']
+        if step['worker_type'] == Step.WorkerType.HUMAN}
+
+    for step_slug, task in iter(project_information['tasks'].items()):
+        # TODO(marcua): Eventually we'll want to support more fine-grained
+        # task filtering for completed vs. in-progress tasks, or to
+        # include machine tasks.
+        if step_slug not in human_slugs:
+            continue
+        if task['status'] != dict(Task.STATUS_CHOICES)[Task.Status.COMPLETE]:
+            continue
+        iterations = []
+        for assignment_idx, assignment in enumerate(task['assignments']):
+            for iteration, snapshot in enumerate(
+                    assignment['snapshots']['snapshots']):
+                if iteration == 0:
+                    iterations.append(Iteration(
+                        project_information['project'], task,
+                        assignment_idx + 1, 0, assignment['worker'],
+                        timedelta(seconds=0),
+                        parse(assignment['start_datetime'])))
+                iterations.append(Iteration(
+                    project_information['project'], task,
+                    assignment_idx + 1,
+                    iteration + 1,
+                    assignment['worker'],
+                    timedelta(seconds=snapshot['work_time_seconds']),
+                    parse(snapshot['datetime'])))
+
+        # To calculate calendar time, zip together each iteration
+        # (in time-order) with the end time of the previous iteration
+        # so that we can subtract the two.
+        iterations = sorted(iterations, key=attrgetter('end_datetime'))
+        iteration_times = (
+            [parse(task['start_datetime'])] +
+            [iteration.end_datetime for iteration in iterations])
+        for start_datetime, iteration in zip(iteration_times[:-1], iterations):
+            yield iteration.to_dict(start_datetime)
--- a/orchestra/management/commands/latency_report.py
+++ b/orchestra/management/commands/latency_report.py
@@ -1,0 +1,30 @@
+"""
+Display a project latency report.
+
+TODO(marcua): Fill the script in.  It's currently a dummy placeholder
+with dummy placeholder arguments.
+"""
+
+from django.core.management.base import BaseCommand
+from orchestra.analytics.latency import work_time_df
+from orchestra.models import Project
+
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def main(options):
+    print(work_time_df(Project.objects.filter(id=1)))
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('--remove-fake',
+                            action='store_true',
+                            dest='remove_fake',
+                            default=False,
+                            help='Remove all fake projects and tasks')
+
+    def handle(self, *args, **options):
+        main(options)
--- a/orchestra/models.py
+++ b/orchestra/models.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from django.contrib.auth.models import User
 from django.db import models
+from django.utils import timezone
 from jsonfield import JSONField
 
 from orchestra.core.errors import ModelSaveError
@@ -9,7 +10,6 @@ from orchestra.workflow import get_step_choices
 from orchestra.workflow import get_workflow_by_slug
 from orchestra.workflow import Step
 from orchestra.utils.assignment_snapshots import load_snapshots
-
 
 # TODO(marcua): Convert ManyToManyFields to django-hstore referencefields or
 # wait for django-postgres ArrayFields in Django 1.8.
@@ -184,6 +184,8 @@ class Task(models.Model):
     A task is a cohesive unit of work representing a workflow step.
 
     Attributes:
+        start_datetime (datetime.datetime):
+            The time the Task was created.
         step_slug (str):
             Identifies the step that the project represents.
         project (orchestra.models.Project):
@@ -209,6 +211,7 @@ class Task(models.Model):
         (Status.ABORTED, 'Aborted'),
         (Status.COMPLETE, 'Complete'))
 
+    start_datetime = models.DateTimeField(default=timezone.now)
     step_slug = models.CharField(max_length=200,
                                  choices=get_step_choices())
     project = models.ForeignKey(Project, related_name='tasks')
@@ -224,7 +227,7 @@ class TaskAssignment(models.Model):
 
     Attributes:
         start_datetime (datetime.datetime):
-            The time the project was created.
+            The time the task was assigned for this iteration.
         worker (orchestra.models.Worker):
             The worker to whom the given task is assigned.
         task (orchestra.models.Task):
@@ -265,7 +268,7 @@ class TaskAssignment(models.Model):
         (Status.PROCESSING, 'Processing'),
         (Status.SUBMITTED, 'Submitted'))
 
-    start_datetime = models.DateTimeField(auto_now_add=True)
+    start_datetime = models.DateTimeField(default=timezone.now)
     worker = models.ForeignKey(Worker,
                                null=True,
                                blank=True)
--- a/orchestra/project_api/api.py
+++ b/orchestra/project_api/api.py
@@ -1,6 +1,8 @@
 from orchestra.models import Project
 from orchestra.workflow import get_workflow_by_slug
+from orchestra.project_api.serializers import ProjectSerializer
 from orchestra.project_api.serializers import TaskSerializer
+
 
 import logging
 
@@ -9,6 +11,19 @@ logger = logging.getLogger(__name__)
 
 class MalformedDependencyException(Exception):
     pass
+
+
+def get_project_information(project_id):
+    project = Project.objects.get(pk=project_id)
+    project_data = ProjectSerializer(project).data
+    tasks = get_project_task_data(project_id)
+    steps = get_workflow_steps(project.workflow_slug)
+
+    return {
+        'project': project_data,
+        'tasks': tasks,
+        'steps': steps
+    }
 
 
 def get_project_task_data(project_id):
@@ -58,8 +73,10 @@ def get_workflow_steps(workflow_slug):
             continue
 
         already_added.add(current_node)
-        steps.append((current_node,
-                      workflow.get_step(current_node).description))
+        current_step = workflow.get_step(current_node)
+        steps.append({'slug': current_node,
+                      'description': current_step.description,
+                      'worker_type': current_step.worker_type})
         for key, dependencies in graph.items():
             if (current_node in dependencies and
                     key not in already_added):
--- a/orchestra/project_api/views.py
+++ b/orchestra/project_api/views.py
@@ -6,11 +6,9 @@ from jsonview.exceptions import BadRequest
 from orchestra.models import Project
 from orchestra.models import WorkerCertification
 from orchestra.project import create_project_with_tasks
-from orchestra.workflow import get_workflows
-from orchestra.project_api.api import get_workflow_steps
-from orchestra.project_api.api import get_project_task_data
+from orchestra.project_api.api import get_project_information
 from orchestra.project_api.decorators import api_endpoint
-from orchestra.project_api.serializers import ProjectSerializer
+from orchestra.workflow import get_workflows
 from urllib.parse import urlparse
 from urllib.parse import urlunsplit
 
@@ -21,22 +19,12 @@ logger = logging.getLogger(__name__)
 @api_endpoint(['POST'])
 def project_information(request):
     try:
-        try:
-            project_id = json.loads(request.body.decode())['project_id']
-        except KeyError:
-            raise BadRequest('project_id is required')
-
-        project = Project.objects.get(pk=project_id)
-        project_data = ProjectSerializer(project).data
-        tasks = get_project_task_data(project_id)
-        steps = get_workflow_steps(project.workflow_slug)
-
-        return {
-            'project': project_data,
-            'tasks': tasks,
-            'steps': steps
-        }
-
+        # TODO(marcua): Add checking for json.loads exceptions to all
+        # endpoints.
+        return get_project_information(
+            json.loads(request.body.decode())['project_id'])
+    except KeyError:
+        raise BadRequest('project_id is required')
     except Project.DoesNotExist:
         raise BadRequest('No project for given id')
 
--- a/orchestra/tests/helpers/fixtures.py
+++ b/orchestra/tests/helpers/fixtures.py
@@ -1,11 +1,14 @@
 import factory
 
+from copy import deepcopy
+from dateutil.parser import parse
 from django.contrib.auth.models import User
 from django.test import Client
 from orchestra.models import Task
 from orchestra.models import TaskAssignment
 from orchestra.models import WorkerCertification
 from orchestra.slack import create_project_slack_group
+from orchestra.utils.task_properties import assignment_history
 
 
 class UserFactory(factory.django.DjangoModelFactory):
@@ -227,7 +230,9 @@ def setup_models(test_case):
     # Create projects
     test_case.projects = {}
     for name, workflow_slug in projects.items():
-        test_case.projects[name] = ProjectFactory(workflow_slug=workflow_slug)
+        test_case.projects[name] = ProjectFactory(
+            workflow_slug=workflow_slug,
+            start_datetime=parse('2015-10-12T00:00:00+00:00'))
         create_project_slack_group(test_case.projects[name])
 
     # Create and assign taks
@@ -235,7 +240,8 @@ def setup_models(test_case):
     for task_slug, details in tasks.items():
         task = TaskFactory(project=test_case.projects[details['project_name']],
                            step_slug=test_case.test_step_slug,
-                           status=details['status'])
+                           status=details['status'],
+                           start_datetime=parse('2015-10-12T01:00:00+00:00'))
         test_case.tasks[task_slug] = task
         for i, (user_id,
                 task_data,
@@ -245,4 +251,56 @@ def setup_models(test_case):
                 task=task,
                 status=assignment_status,
                 assignment_counter=i,
-                in_progress_task_data=task_data)
+                in_progress_task_data=task_data,
+                start_datetime=parse(
+                    '2015-10-12T0{}:00:00+00:00'.format(2 + i)))
+
+
+def setup_task_history(test):
+    task = test.tasks['rejected_review']
+
+    test._submit_assignment(
+        test.clients[6], task.id, seconds=35)
+    test._submit_assignment(
+        test.clients[7], task.id, command='reject', seconds=36)
+    test._submit_assignment(
+        test.clients[6], task.id, seconds=37)
+    test._submit_assignment(
+        test.clients[7], task.id, command='accept', seconds=38)
+
+    # Fill out the snapshots for all assignments
+    assignments = assignment_history(task)
+    first_assignment = assignments[0]
+    second_assignment = assignments[1]
+    third_assignment = assignments[2]
+
+    first_assignment.snapshots['snapshots'] = deepcopy(
+        second_assignment.snapshots['snapshots'])
+    second_assignment.snapshots['snapshots'] = (
+        deepcopy(third_assignment.snapshots['snapshots']) +
+        second_assignment.snapshots['snapshots'])[:-1]
+
+    def fix_datetimes(snapshots, new_datetimes):
+        for snapshot, new_datetime in zip(snapshots, new_datetimes):
+            snapshot['datetime'] = new_datetime
+
+    # Explicitly set the iteration datetimes.  If we didn't, the timestamps
+    # would be `datetime.now`, which we can't test against.  The explicitly set
+    # times are predictable distance apart, so we can test the
+    # resulting latency reports.
+    fix_datetimes(
+        first_assignment.snapshots['snapshots'],
+        ['2015-10-12T02:02:00+00:00', '2015-10-12T03:05:00+00:00'])
+    fix_datetimes(
+        second_assignment.snapshots['snapshots'],
+        ['2015-10-12T03:01:00+00:00', '2015-10-12T03:07:00+00:00',
+         '2015-10-12T04:03:00+00:00', '2015-10-12T04:10:00+00:00'])
+    fix_datetimes(
+        third_assignment.snapshots['snapshots'],
+        ['2015-10-12T04:02:00+00:00', '2015-10-12T04:13:00+00:00'])
+
+    first_assignment.save()
+    second_assignment.save()
+    third_assignment.save()
+
+    return task
--- a/orchestra/tests/test_analytics.py
+++ b/orchestra/tests/test_analytics.py
@@ -1,0 +1,39 @@
+from datetime import timedelta
+from dateutil.parser import parse
+from orchestra.analytics.latency import work_time_df
+from orchestra.tests.helpers import OrchestraTestCase
+from orchestra.tests.helpers.fixtures import setup_models
+from orchestra.tests.helpers.fixtures import setup_task_history
+from pandas import DataFrame
+from pandas.util.testing import assert_frame_equal
+
+
+class AnalyticsTestCase(OrchestraTestCase):
+    def setUp(self):  # noqa
+        super(AnalyticsTestCase, self).setUp()
+        setup_models(self)
+
+    def test_work_time_df(self):
+        task = setup_task_history(self)
+        returned = work_time_df([task.project])
+
+        self.assertEquals(returned.project_id.nunique(), 1)
+        self.assertEquals(returned.task_id.nunique(), 1)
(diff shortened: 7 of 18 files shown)


orchestra/models.py
```diff
@@ -265,7 +268,7 @@ class Status:
         (Status.PROCESSING, 'Processing'),
         (Status.SUBMITTED, 'Submitted'))
 
-    start_datetime = models.DateTimeField(auto_now_add=True)
+    start_datetime = models.DateTimeField(default=timezone.now)
```
marcua, line 271: open question: should we set current values to `None` and write some sort of migration script, or just move on with some dirty data for old tasks' `start_datetime`?

orchestra/project_api/api.py
```diff
@@ -58,8 +73,10 @@ def get_workflow_steps(workflow_slug):
             continue
 
         already_added.add(current_node)
-        steps.append((current_node,
-                      workflow.get_step(current_node).description))
+        current_step = workflow.get_step(current_node)
+        steps.append({'slug': current_node,
```
marcua, line 77: changed API to dict of step information rather than a tuple, since we went from a pair to a triple

orchestra/project_api/api.py
```diff
@@ -58,8 +73,10 @@ def get_workflow_steps(workflow_slug):
...
+                      'description': current_step.description,
+                      'worker_type': current_step.worker_type})
```
marcua, line 79: need to filter to only human tasks for latency reporting

orchestra/project_api/api.py
```diff
@@ -11,6 +13,19 @@ class MalformedDependencyException(Exception):
     pass
 
 
+def get_project_information(project_id):
```
marcua, line 16: refactored this so it could get used both by the API and by latency reporting

orchestra/tests/helpers/fixtures.py
```diff
@@ -245,4 +251,52 @@ def setup_models(test_case):
                 task=task,
                 status=assignment_status,
                 assignment_counter=i,
-                in_progress_task_data=task_data)
+                in_progress_task_data=task_data,
+                start_datetime=parse(
+                    '2015-10-12T0{}:00:00+00:00'.format(2 + i)))
+
+
+def task_back_and_forth(test):
```
marcua: in the next iteration, i'll change the name of this thing and add a docstring.  roughly: creates the back-and-forth assignment iteration and timing information so we can test latency reporting, etc.

orchestra/tests/test_dashboard.py
```diff
@@ -470,11 +472,12 @@ def test_assignment_snapshot_upgrades(self):
         good_snapshots = {
             'snapshots': [{
                 'data': {},
-                'datetime': datetime.utcnow().isoformat(),
+                'datetime': (
+                    datetime.utcnow().replace(tzinfo=tzutc()).isoformat()),
```
marcua: we were timezone-naive before :(

orchestra/tests/test_dashboard.py
```diff
@@ -574,16 +567,3 @@ def _verify_good_task_assignment_information(
             }
         }
         self.assertEquals(returned, expected)
-
-    def _submit_assignment(self, client, task_id, data=None,
```
marcua, line 578: moved this into the base OrchestraTestCase class so that it could be used in latency reporting

marcua:
I apologize for the additional commits on top of the first one---i left explanatory comments on the code that i don't want the reviewer to lose if I `git push -f`

orchestra/models.py
thisisdhaas, line 271: why would the old ones be dirty? Does auto_now_add not add timezone-aware datetimes?

orchestra/project_api/api.py
thisisdhaas, line 77: okay--worth a minor version bump?

orchestra/project_api/api.py
thisisdhaas, line 79: :white_check_mark:

orchestra/project_api/api.py
thisisdhaas, line 16: :white_check_mark:

orchestra/tests/test_dashboard.py
thisisdhaas: Can we use something from `django.util.timezone` here? I get nervous about doing timezones right in python...

orchestra/tests/test_dashboard.py
thisisdhaas, line 578: :cherries: = :white_check_mark:

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+from orchestra.models import Task
+from orchestra.project_api.api import get_project_information
+from orchestra.workflow import Step
+from pandas import DataFrame
+
+
+def work_time_df(projects):
+    """
+    Return all projects' task assignment iteration timing information as a
+    dataframe.
```
thisisdhaas: docstring nit: keep the docstring summary to one line.

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+
+    The table schema is:
```
thisisdhaas: Is there a better word for dataframe schema? Is it `dataframe schema`?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        project: the project's ID
```
thisisdhaas: `project_id`?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        task: the task's ID
```
thisisdhaas: `task_id`?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        assignment: a 0-indexed assignment counter (0 is entry-level, 1 is
+            first review, etc.)
```
thisisdhaas: make 1-indexed to clearly separate from the semantics of `iteration 0`.

Also, maybe this should be `assignment_level`, since `assignment 1` feels like 'the first worker to work on the task', not 'the worker at the first level of the hierarchy working on the task'.

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        iteration: the work iteration for that assignment.
+            - assignment 0 iteration 0 represents the period from task creation
+              to first assignment.
+            - assignment N > 0 iteration 0 represents the period from previous
+              submission to reviewer assignment.
```
thisisdhaas: clarify this documentation:
- iteration 0 represents the period from previous submission to reviewer assignment:
  - If this is the first assignment, 'previous submission' means 'task creation'.
  - Work time will always be 0
  - If this task is auto-assigned, calendar time will also be 0.
- Iterations N > 0 represent the period between worker assignment and worker submission.

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+            that iteration (0 for iteration 0).
+
+    Args:
+        projects (django.db.models.query.QuerySet):
+            An iterable of projects to include in the report.
+
+    Returns:
+        df (pandas.DataFrame):
+            A DataFrame with timing information.
+    """
```
thisisdhaas, line 44: Consider also adding `worker_id` to enable worker-based analysis (only suggesting it because it feels easy).

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+    # TODO(marcua): `get_project_information` per project means we do random
+    # access all over the database.  Ideally, we'd have a sequential scan
+    # version of this.
```
thisisdhaas, line 47: or even a single glorious query! (c.f. [select_related](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#select-related) or [prefecth_related](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#prefetch-related)) I guess that would need a `get_projects_information()` call.

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+    time_data = list(itertools.chain.from_iterable(
+        project_time_row_generator(
+            get_project_information(project.id))
+        for project in projects))
+    df = DataFrame(time_data)
+    df = df[
+        ['project', 'task', 'assignment', 'iteration', 'start_datetime',
+         'calendar_time', 'work_time']]
```
thisisdhaas: why is this projection necessary?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+    # lol at this variable name
+    human_slugs = {
+        step['slug'] for step in project_information['steps']
+        if step['worker_type'] == Step.WorkerType.HUMAN}
+
+    for step_slug, task in iter(project_information['tasks'].items()):
+        if step_slug not in human_slugs:
+            continue
+        if task['status'] != dict(Task.STATUS_CHOICES)[Task.Status.COMPLETE]:
+            continue
```
thisisdhaas, line 98: Add a TODO to support filtering by these things one day?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        iterations = []
+        for assignment_idx, assignment in enumerate(task['assignments']):
+            for iteration, snapshot in enumerate(
+                    assignment['snapshots']['snapshots']):
```
thisisdhaas, line 102: this ugliness comes from the way we serialize `TaskAssignment`s?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+                if iteration == 0:
+                    iterations.append(Iteration(
+                        project_information['project'], task,
+                        assignment_idx, 0, timedelta(seconds=0),
```
thisisdhaas: `assignment_idx+1`

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+                        parse(assignment['start_datetime'])))
```
thisisdhaas, line 108: so we're using Django Rest Framework to serialize datetimes as strings, then re-parsing them as datetimes? Is there a cleaner way to do this?

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+                iterations.append(Iteration(
+                    project_information['project'], task,
+                    assignment_idx,
```
thisisdhaas: `assignment_idx+1`

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+                    iteration + 1,
+                    timedelta(seconds=snapshot['work_time_seconds']),
+                    parse(snapshot['datetime'])))
+        iterations = sorted(iterations, key=attrgetter('end_datetime'))
```
thisisdhaas, line 120: why? shouldn't we be sorting the output dataframe if we want the 'order by time' semantics? I don't see why this is better than sorting by (assignment, iteration).

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+        iteration_times = (
+            [parse(task['start_datetime'])] +
+            [iteration.end_datetime for iteration in iterations])
+        for start_datetime, iteration in zip(iteration_times[:-1], iterations):
```
thisisdhaas: couldn't this just be:

    start_datetime = parse(task['start_datetime'])
    for iteration in iterations:
        yield iteration.to_dict(start_datetime)

orchestra/analytics/latency.py
```diff
@@ -0,0 +1,112 @@
...
+            yield iteration.to_dict(start_datetime)
```
thisisdhaas, line 125: ah I see why you're sorting. Do we really have no other way to get `start_datetime` correctly from the snapshot?

orchestra/models.py
```diff
@@ -10,7 +11,6 @@
 from orchestra.workflow import Step
 from orchestra.utils.assignment_snapshots import load_snapshots
 
-
 # TODO(marcua): Convert ManyToManyFields to django-hstore referencefields or
 # wait for django-postgres ArrayFields in Django 1.8.
```
thisisdhaas, line 15: not that this is relevant to your PR, but why would we want to do this?

orchestra/project_api/views.py
```diff
@@ -21,22 +19,10 @@
...
-        steps = get_workflow_steps(project.workflow_slug)
-
-        return {
-            'project': project_data,
-            'tasks': tasks,
-            'steps': steps
-        }
-
+        return get_project_information(
+            json.loads(request.body.decode())['project_id'])
```
thisisdhaas, line 25: handle the JSON parse error as well to avoid 500s?

orchestra/tests/helpers/fixtures.py
```diff
@@ -235,7 +240,8 @@ def setup_models(test_case):
     for task_slug, details in tasks.items():
         task = TaskFactory(project=test_case.projects[details['project_name']],
                            step_slug=test_case.test_step_slug,
-                           status=details['status'])
+                           status=details['status'],
+                           start_datetime=parse('2015-10-12T01:00:00+00:00'))
```
thisisdhaas, line 244: why not just create a datetime object?

orchestra/tests/helpers/fixtures.py
```diff
@@ -227,7 +230,9 @@ def setup_models(test_case):
     # Create projects
     test_case.projects = {}
     for name, workflow_slug in projects.items():
-        test_case.projects[name] = ProjectFactory(workflow_slug=workflow_slug)
+        test_case.projects[name] = ProjectFactory(
+            workflow_slug=workflow_slug,
+            start_datetime=parse('2015-10-12T00:00:00+00:00'))
```
thisisdhaas, line 235: why not just create a datetime object?

orchestra/tests/helpers/fixtures.py
```diff
@@ -245,4 +251,52 @@ def setup_models(test_case):
                 task=task,
                 status=assignment_status,
                 assignment_counter=i,
-                in_progress_task_data=task_data)
+                in_progress_task_data=task_data,
+                start_datetime=parse(
+                    '2015-10-12T0{}:00:00+00:00'.format(2 + i)))
```
thisisdhaas, line 256: why not just create a datetime object?

orchestra/tests/helpers/fixtures.py
```diff
@@ -245,4 +251,52 @@ def setup_models(test_case):
...
+    test._submit_assignment(
+        test.clients[6], task.id, seconds=35)
+    test._submit_assignment(
+        test.clients[7], task.id, command='reject', seconds=36)
+    test._submit_assignment(
+        test.clients[6], task.id, seconds=37)
+    test._submit_assignment(
+        test.clients[7], task.id, command='accept', seconds=38)
+
+    # Fill out the snapshots for all assignments
```
thisisdhaas, line 271: perhaps this should be pulled out as a helper method?

orchestra/tests/helpers/fixtures.py
```diff
@@ -245,4 +251,52 @@ def setup_models(test_case):
...
+        first_assignment.snapshots['snapshots'],
+        ['2015-10-12T02:02:00+00:00', '2015-10-12T03:05:00+00:00'])
+    fix_datetimes(
+        second_assignment.snapshots['snapshots'],
+        ['2015-10-12T03:01:00+00:00', '2015-10-12T03:07:00+00:00',
+         '2015-10-12T04:03:00+00:00', '2015-10-12T04:10:00+00:00'])
+    fix_datetimes(
+        third_assignment.snapshots['snapshots'],
+        ['2015-10-12T04:02:00+00:00', '2015-10-12T04:13:00+00:00'])
+
```
thisisdhaas, line 301: I didn't understand why all of this datetime fixing was necessary.

orchestra/tests/helpers/fixtures.py
thisisdhaas: Perhaps `setup_task_history()`?

orchestra/tests/helpers/fixtures.py
```diff
@@ -245,4 +251,52 @@ def setup_models(test_case):
...
                 status=assignment_status,
                 assignment_counter=i,
-                in_progress_task_data=task_data)
+                in_progress_task_data=task_data,
+                start_datetime=parse(
+                    '2015-10-12T0{}:00:00+00:00'.format(2 + i)))
+
+
+def task_back_and_forth(test):
+    task = test.tasks['rejected_review']
```
thisisdhaas, line 260: should the function take `task_slug` as an argument, or will we always use `'rejected_review'`?

orchestra/tests/test_analytics.py
```diff
@@ -0,0 +1,39 @@
...
+from orchestra.analytics.latency import work_time_df
+from orchestra.tests.helpers import OrchestraTestCase
+from orchestra.tests.helpers.fixtures import setup_models
+from orchestra.tests.helpers.fixtures import task_back_and_forth
+from pandas import DataFrame
+from pandas.util.testing import assert_frame_equal
+
+
+class AnalyticsTestCase(OrchestraTestCase):
+    def setUp(self):  # noqa
```
thisisdhaas, line 12: Out of curiosity, what's the noqa?

orchestra/tests/test_dashboard.py
```diff
@@ -470,11 +472,12 @@ def test_assignment_snapshot_upgrades(self):
...
             'snapshots': [{
                 'data': {},
-                'datetime': datetime.utcnow().isoformat(),
+                'datetime': (
+                    datetime.utcnow().replace(tzinfo=tzutc()).isoformat()),
                 'type': TaskAssignment.SnapshotType.SUBMIT,
                 'work_time_seconds': 1
             }],
-            '__version': 1}
+            '__version': 2}
```
thisisdhaas, line 479: what version is this?

orchestra/tests/test_dashboard.py
```diff
@@ -483,13 +486,15 @@ def test_assignment_snapshot_upgrades(self):
                     'type': TaskAssignment.SnapshotType.SUBMIT,
                 }, {
                     'data': {},
-                    'datetime': datetime.utcnow().isoformat(),
+                    'datetime': (
+                        datetime.utcnow().replace(tzinfo=tzutc()).isoformat()),
```
thisisdhaas: again, something from util.timezone?

orchestra/tests/test_dashboard.py
```diff
@@ -483,13 +486,15 @@ def test_assignment_snapshot_upgrades(self):
...
                     'type': TaskAssignment.SnapshotType.SUBMIT,
                     'work_time_seconds': 1
                 }]}
         upgraded_snapshots = deepcopy(old_snapshots)
-        upgraded_snapshots['__version'] = 1
+        upgraded_snapshots['__version'] = 2
         upgraded_snapshots['snapshots'][0]['work_time_seconds'] = 0
+        upgraded_snapshots['snapshots'][0]['datetime'] += '+00:00'
```
thisisdhaas, line 496: Seems like we should be serializing the datetime with timezone information if we're just going to append it anyways, is there a way to do that?

orchestra/utils/assignment_snapshots.py
```diff
@@ -26,7 +29,14 @@ def load_snapshots(snapshots):
...
         snapshots[SNAPSHOTS] = snapshot_list
         snapshots[VERSION] = 1
-    if snapshots[VERSION] != 1:
+    if snapshots[VERSION] == 1:
+        snapshot_list = snapshots.get(SNAPSHOTS, [])
+        for snapshot_data in snapshot_list:
+            snapshot_data[DATETIME] = parse(
+                snapshot_data[DATETIME]).replace(tzinfo=tzutc()).isoformat()
+        snapshots[SNAPSHOTS] = snapshot_list
+        snapshots[VERSION] = 2
```
thisisdhaas, line 38: so the goal is to support old snapshots at varying versions? Can we at least pull this out into a `migrate_snapshot(from, to)` method? Also maybe call out the latest version somewhere?

orchestra/utils/task_lifecycle.py
```diff
@@ -773,7 +774,7 @@ def submit_task(task_id, task_data, snapshot_type, worker, work_time_seconds):
     assignment.in_progress_task_data = task_data
     assignment.snapshots['snapshots'].append(
         {'data': assignment.in_progress_task_data,
-         'datetime': datetime.utcnow().isoformat(),
+         'datetime': datetime.utcnow().replace(tzinfo=tzutc()).isoformat(),
```
thisisdhaas: the usual biznaz about django.util.timezone

orchestra/models.py
marcua, line 271: oh woops---this comment applies to the new field on Task---what should we do to old Tasks?

orchestra/project_api/api.py
marcua, line 77: i'd bump the third number in the triple for sure :)

orchestra/tests/test_dashboard.py
marcua: this is JSON and not django---do you still think we should use django tooling for it?

orchestra/analytics/latency.py
marcua: reorders the columns---mostly for pretty printing

orchestra/analytics/latency.py
marcua, line 102: yea so snapshots is a field that requires a jsonfield, which is itself a dict :/

orchestra/analytics/latency.py
marcua, line 108: but it's on opposite sides of the client/server boundary, so that feels right, no?

orchestra/analytics/latency.py
marcua, line 120: this is a trick to get the times and zip(start_times, iterations) in the following lines---is this worth a comment?

orchestra/analytics/latency.py
marcua, line 125: i couldn't think of a way to do it given that you might have different reject/accept/submit patterns up and down the hierarchy.  can you?

orchestra/models.py
marcua, line 15: if we use arrayfields, we can push operations like filter/project into postgres even for JSON fields---the data is no longer opaque.  worth updating the comment?

orchestra/project_api/views.py
marcua, line 25: were we doing that before, or is this a new thing?

orchestra/tests/helpers/fixtures.py
marcua, line 244: this seemed more compact to include the timezone---is that alright?

orchestra/tests/helpers/fixtures.py
marcua, line 271: i like to refactor once there's a second use-case.  thoughts?

orchestra/tests/helpers/fixtures.py
marcua, line 301: otherwise the timestamp is datetime.now, which we can't test against.  these times are predictable distance apart, so I can test the resulting dataframe.  worth a comment?

orchestra/tests/helpers/fixtures.py
marcua, line 260: yea for now it only has 1.5 use cases :)

orchestra/tests/test_analytics.py
marcua, line 12: the name `setUp` breaks pep8, but setUp is the name TestCase requires

orchestra/tests/test_dashboard.py
marcua, line 479: the version helps us do migrations on the data.  see `orchestra/utils/assignment_snapshots.py`

orchestra/tests/test_dashboard.py
marcua: i'll do whatever based on what you reply to the other comment

orchestra/tests/test_dashboard.py
marcua, line 496: i'll do whatever based on what you reply to the other comment

orchestra/utils/assignment_snapshots.py
marcua, line 38: the entire purpose of this function is as a migration.  i'm happy to add a docstring, but i don't believe the new function is necessary.

orchestra/models.py
thisisdhaas, line 271: ah--yeah the default would create them all with a timestamp of whenever you run the migration. I'm not sure what the correct value would even be, here's one proposal:
- migration 1 adds the field, setting all values to the present datetime
- data migration 2 changes all `task.start_datetime` to `task.assignments.first().start_datetime`

orchestra/tests/test_dashboard.py
thisisdhaas: At the very least `datetime.utcnow().replace(tzinfo=tzutc()).isoformat()` => `django.util.timezone.now().isoformat()`.

The django utils are nice wrappers that allow us not to think about timezones, I'm for using them whenever we can.

orchestra/analytics/latency.py
thisisdhaas: Ah, ok. Should that happen in the analysis script? (feel free to say no and leave it be).

orchestra/analytics/latency.py
thisisdhaas, line 102: yay abstraction

orchestra/analytics/latency.py
thisisdhaas, line 108: oh it is? I though you were using the server-side 'get_project_information', not the client API call.

orchestra/analytics/latency.py
thisisdhaas, line 120: maybe a comment for the remainder of the function, yeah.

orchestra/analytics/latency.py
thisisdhaas, line 125: Not off the top of my head, feel free to leave as is.

orchestra/models.py
thisisdhaas, line 15: eh, as long as you know what it means. But a lot of our ManyToManyFields are legitimate, right? (e.g. the self-reference in Certifications.required_certifications).

orchestra/project_api/views.py
thisisdhaas, line 25: I guess it would be new. That a bad thing?

orchestra/tests/helpers/fixtures.py
thisisdhaas, line 244: https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.timezone.make_aware would be my preference (no argument necessary for `timezone`, it will be pulled from settings as UTC) :-)

But feel free to overrule if you don't like my django-utils-fanboyism.

orchestra/tests/helpers/fixtures.py
thisisdhaas, line 271: yeah, wrote this comment before I realized the method wasn't inside an individual test. Feel free to leave as is.

orchestra/tests/helpers/fixtures.py
thisisdhaas, line 301: yep, comment makes me happy.

orchestra/tests/helpers/fixtures.py
thisisdhaas, line 260: :cactus: = :white_check_mark:

orchestra/tests/test_analytics.py
thisisdhaas, line 12: silly pep8, trix are for kids.

orchestra/tests/test_dashboard.py
thisisdhaas, line 496: I think that a timezone-aware datetime will output the '+offset' with `isoformat()`, but that's worth checking. If not, I guess this is fine.

orchestra/utils/assignment_snapshots.py
thisisdhaas, line 38: okay, fine by me.

orchestra/tests/test_dashboard.py
marcua: :+1: ty!

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua: done, with a small number of corrections

orchestra/analytics/latency.py
marcua, line 44: worker_id is harder because it's not in the API.  worker_name was, and so i added that.  for this PR, I'm either willing to revert that addition (because worker_id is cleaner) or leave it with my new addition.

orchestra/analytics/latency.py
marcua, line 47: agree

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua, line 98: done

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua, line 108: and that convenience method is doing the serialization for the API.

changing that would require two serialization phases, which is way worse than having to deserialize imho

orchestra/analytics/latency.py
marcua: done

orchestra/analytics/latency.py
marcua, line 120: done

orchestra/models.py
marcua, line 15: oh sorry this was different---it's just to reduce joins when possible.  anyway :)

orchestra/tests/helpers/fixtures.py
marcua, line 244: I don't like that it will be pulled from settings as whatever is in settings, which is implicit rather than explicit :).  leaving it explicit.

orchestra/tests/helpers/fixtures.py
marcua: done

orchestra/tests/helpers/fixtures.py
marcua, line 301: done

orchestra/tests/test_dashboard.py
marcua, line 496: `upgraded_snapshots` is me using an alternative method of verifying that the upgrade logic worked, rather than using the same code as the implementation :)

orchestra/models.py
marcua, line 271: did that, but start_datetime is task.project.start_datetime.  lmk if that works

orchestra/project_api/views.py
marcua, line 25: added a TODO, since we have to do it for all endpoints

orchestra/tests/helpers/fixtures.py
marcua, line 235: kept consistent with the other ones

orchestra/tests/helpers/fixtures.py
marcua, line 256: my consistent aversion to including lots-o timezones :)

orchestra/tests/test_analytics.py
marcua, line 12: :)

orchestra/utils/task_lifecycle.py
marcua: done-zo

orchestra/models.py
thisisdhaas, line 271: :santa:

orchestra/project_api/api.py
thisisdhaas, line 77: maybe using an exciting new script (shameless plug #53)

orchestra/analytics/latency.py
thisisdhaas, line 44: I'm fine with that--should still be useful for analysis. Add a TODO to change to worker_id if we update the API?

orchestra/analytics/latency.py
thisisdhaas, line 108: I'm convinced.

orchestra/tests/helpers/fixtures.py
thisisdhaas: hm, I'd still rather see `timezone.makeaware(datetime(year=2015, month=10, day=12, hour=1), timezone.utc)`, but I'm getting really tired of trying to care about timestamps, so also feel free to leave it.

orchestra/management/commands/latency_report.py
```diff
@@ -0,0 +1,27 @@
...
+logger = logging.getLogger(__name__)
+
+
+def main(options):
+    print(work_time_df(Project.objects.filter(id=1)))
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('--remove-fake',
```
thisisdhaas, line 23: add a comment: this is an incomplete script, and a dummy example argument.

thisisdhaas:
LGTM once you add the comment I asked for. latency is awesome

PR #51, opened 2015-10-13
Separate styles by feature

jrbotros:
To start to cleanup our scsspaghetti and allow components to use common variables/mixins. See #52 for future enhancements.


marcua:
it's not totally clear to me when to import common and when to import base.  can you include a comment in each to that effect, and then you can merge w/o another review?

---

@jrbotros
:white_check_mark:

PR #54, opened 2015-10-15
Workflows near db

thisisdhaas:
Moves workflows from code objects to django apps that get loaded into the database.

Includes:
- example app for the simple_workflow with correct json manifests
- management command to load an app as a workflow
- new database models for workflows, workflow versions, and workflow steps
- updated models for objects that relate to the new models
- migrations for the models (and a data migration to upgrade the data)
- modified existing codebase to use new database models
- modified existing tests to use new database models
- ported the journalism workflow to the new app format
- added tests for the new code
- verified that the admin still works with the new models.
- updated docs for the new code
- Added sample data loading mechanism for workflows.


orchestra/management/commands/loadworkflow.py
```diff
@@ -0,0 +1,35 @@
+from django.core.management.base import BaseCommand, CommandError
```
thisisdhaas: this is the main script for loading a django app into the db as a workflow.

orchestra/migrations/0017_auto_20151007_1424.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
```
thisisdhaas, line 1: this migration adds new models and changes existing models in a nullable/non-breaking way.

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
```
thisisdhaas, line 1: This is the data migration that will handle existing projects and tasks.

It should only be run _after_ the new workflows are loaded into the DB.

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
...
+from django.db import models, migrations
+
+
+def update_projects(apps, schema_editor):
+    Project = apps.get_model('orchestra', 'Project')
+    WorkflowVersion = apps.get_model('orchestra', 'WorkflowVersion')
+
+    for project in Project.objects.all():
+        project.workflow_version = WorkflowVersion.objects.get(
+            slug=project.workflow_slug)
```
thisisdhaas, line 28: This assumes that when an existing workflow is implemented with a json manifest, the version_slug matches the old workflow_slug.

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
...
+    Task = apps.get_model('orchestra', 'Task')
+    WorkflowStep = apps.get_model('orchestra', 'WorkflowStep')
+
+    for task in Task.objects.all():
+        task.step = WorkflowStep.objects.get(slug=task.step_slug)
+        task.save()
+
+
+def migrate_worker_certs(apps, schema_editor):
+    Certification = apps.get_model('orchestra', 'Certification')
```
thisisdhaas, line 42: This handles the fact that certifications didn't used to be scoped to a workflow. This function takes existing worker_certifications and creates a copy for every new workflow that defines a certifications with the same slug.

orchestra/migrations/0019_auto_20151007_1846.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
```
thisisdhaas, line 1: This migration drops old fields and makes the friendly nullable fields non-nullable once the data is all in place.

orchestra/models.py
```diff
@@ -14,6 +12,56 @@
...
+            The full path to the location of the workflow's manifest.
+    """
+    slug = models.CharField(max_length=200, unique=True)
+    name = models.CharField(max_length=200)
+    description = models.TextField()
+    code_directory = models.CharField(max_length=255, unique=True)
+
+    class Meta:
+        app_label = 'orchestra'
+
```
thisisdhaas, line 39: A future update will add `__str__()` methods to these classes.

orchestra/models.py
```diff
@@ -14,6 +12,56 @@
...
+            Human-readable name for the workflow.
+        description (str):
+            A longer description of the workflow.
+        code_directory (str):
+            The full path to the location of the workflow's manifest.
+    """
+    slug = models.CharField(max_length=200, unique=True)
+    name = models.CharField(max_length=200)
+    description = models.TextField()
+    code_directory = models.CharField(max_length=255, unique=True)
```
thisisdhaas, line 34: This uniqueness means that you can't register the same app (workflow) twice in the DB.

orchestra/models.py
```diff
@@ -28,16 +76,92 @@ class Certification(models.Model):
...
+        related_name='dependent_certifications',
+        symmetrical=False)
+    workflow = models.ForeignKey(Workflow, related_name='certifications')
 
     def __str__(self):
         return '{}'.format(self.slug)
 
+    class Meta:
+        app_label = 'orchestra'
+        unique_together = (('workflow', 'slug'),)
```
thisisdhaas, line 102: this (and the `workflow` field) scopes certifications to workflows.

orchestra/workflow/__init__.py
```diff
@@ -0,0 +1 @@
+from orchestra.workflow.workflow import *  # noqa
```
thisisdhaas: This line (and `orchestra.workflow.workflow`) go away in the future.

orchestra/workflow/directory.py
```diff
@@ -0,0 +1,36 @@
+import json
```
thisisdhaas, line 1: Implementation of parsing workflows from directories containing manifests.

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
+from django.db import transaction
```
thisisdhaas, line 2: implementation for loading parsed directories into the Django database.

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        create_or_update(
+            Certification,
+            {
+                'slug': certification_data['slug'],
+                'workflow': workflow
+            },
+            {
+                'name': certification_data['name'],
+                'description': certification_data['description'],
+            }
```
thisisdhaas, line 40: This (and the block above) avoid errors with loading different versions at different times (e.g., loading v1 creates the workflow, loading v2 shouldn't complain because the workflow exists). It also updates the workflow and its certifications based on the latest contents of the manifest.

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        required_cert_slugs = certification_data['required_certifications']
+        required_certs = Certification.objects.filter(
+            workflow=workflow,
+            slug__in=required_cert_slugs
+        )
+        if required_certs.count() != len(required_cert_slugs):
+            raise WorkflowError(
+                'Certification {} requires non-existent certification.'
+                .format(certification_data['slug']))
+        cert.required_certifications = list(required_certs)
```
thisisdhaas: This doesn't error if certification dependencies change over time, which might be the wrong move. Thoughts? Since we don't version certifications, I don't know what the right move is here.

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        },
+        {
+            'name': version_data['name'],
+            'description': version_data['description']
+        }
+    )
+
+    if not version_created:
+        if not force:
+            raise WorkflowError('Version {} already exists'
```
thisisdhaas, line 89: unless `--force` was passed, don't allow versions to be updated. We're allowed to raise an exception after creating the object because this is all wrapped in a single transaction.

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        old_step_slugs = set(
+            WorkflowStep.objects
+            .filter(workflow_version=version)
+            .values_list('slug', flat=True)
+        )
+        if new_step_slugs != old_step_slugs:
+            raise WorkflowError('Even with --force, cannot change the steps '
+                                'of a workflow. Drop and recreate the '
+                                'database to reset, or create a new version '
+                                'for your workflow.')
```
thisisdhaas, line 103: With `--force`, this checks that the step slugs haven't changed. Are there other things that need to be checked here? (see my TODO below for an example).

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        required_certification_slugs = step_data.get(
+            'required_certifications', [])
+        required_certifications = list(Certification.objects.filter(
+            workflow=workflow,
+            slug__in=required_certification_slugs
+        ))
+        step.required_certifications = required_certifications
+
+    # Set up step dependencies once the steps objects are in the DB.
+    # TODO(dhaas): should we prevent updates to these?
```
thisisdhaas: It seems like we should definitely prevent these updates. But I'm not sure how to do so easily.

orchestra/workflow/manifest.py
```diff
@@ -0,0 +1,6 @@
+def parse_version_manifest(manifest):
```
thisisdhaas: This file will be gone in the future.

orchestra/management/commands/loadworkflow.py
marcua: docstring to that effect :)

orchestra/management/commands/loadworkflow.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Loads a workflow version into the database'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'workflow_directory',
+            help='The code directory containing the workflow.')
+        parser.add_argument(
```
marcua, line 13: feels like it should be required

orchestra/management/commands/loadworkflow.py
```diff
@@ -0,0 +1,35 @@
...
+            'version_slug', nargs='?', default=None,
+            help=('(Optional) The version to load. If omitted, all available '
+                  'versions are loaded.'))
+        parser.add_argument(
+            '--force', action='store_true', default=False,
+            help=('If a workflow version already exists, update the existing '
+                  'version instead of exiting.'))
+
+    def handle(self, *args, **options):
+        workflow_dir = os.path.abspath(options['workflow_directory'])
```
marcua: it'd be nice if this called code in `orchestra` for any business logic, so the script was just the wrapper around that code

orchestra/management/commands/loadworkflow.py
```diff
@@ -0,0 +1,35 @@
...
+        if not os.path.exists(workflow_dir):
+            raise CommandError('Workflow directory does not exist.')
+
+        version_slug = options['version_slug']
+        versions = version_slug if version_slug is None else [version_slug]
```
marcua: check if version exists in directory?

orchestra/migrations/0018_auto_20151007_1451.py
marcua, line 1: high-level: i worry about how automatic the three migrations are.  if someone forgets to add their workflows to the filesystem, for example we might kill a bunch of data and it wont' be recoverable.  any way to avoid this, perhaps by having people explicitly call each step?

oh right: i'm vaguely remembering that we said these WOULD be manual.  should we include a comment here to that effect?

orchestra/management/commands/loadworkflow.py
derek-schultz: Do we still import one per line?

orchestra/models.py
marcua, line 34: since the slug and code directory are separate, are we sure we benefit from uniqueness somehow?  seems like we might regret it down the line?

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
...
+    WorkerCertification = apps.get_model('orchestra', 'WorkerCertification')
+
+    # Copy unscoped worker certifications to the new scoped objects.
+    unscoped_worker_certs = WorkerCertification.objects.filter(
```
marcua: general style: spell out `certs`

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
...
+                certification=cert,
+                worker=worker_cert.worker,
+                task_class=worker_cert.task_class,
+                role=worker_cert.role
+            )
+
+        # Delete worker certs we successfully copied.
+        # Necessary so we can make Certification.workflow non-nullable.
+        if scoped_certs.exists():
+            worker_cert.delete()
```
marcua: no weird cascades, etc?
are there any weird semantics on delete with a null in the filter?

scary :)

orchestra/migrations/0018_auto_20151007_1451.py
```diff
@@ -0,0 +1,78 @@
...
+    # Verify that we've migrated all unscoped worker certifications
+    unmigrated_certifications = (Certification.objects
+                                 .exclude(workercertification=None)
+                                 .filter(workflow__isnull=True))
+    if unmigrated_certifications.exists():
+        raise ValueError('Failed to migrate all worker certifications!')
+
+    # Delete unscoped certifications.
+    # Necessary so we can make Certification.workflow non-nullable.
+    Certification.objects.filter(workflow__isnull=True).delete()
```
marcua, line 79: do we know of databases for which this is true? :)

orchestra/models.py
```diff
@@ -14,6 +12,56 @@
 # TODO(marcua): Convert ManyToManyFields to django-hstore referencefields or
 # wait for django-postgres ArrayFields in Django 1.8.
 
+class Workflow(models.Model):
+    """
+    Workflows describe the steps and requirements of completing expert work.
```
marcua: "for experts to complete work."

orchestra/models.py
```diff
@@ -14,6 +12,56 @@
...
+
+    This model represents workflows that have been loaded into the database and
+    are usable by Orchestra projects. All workflows are also defined in the
+    codebase (see the `root_directory` attribute).
```
marcua: code_directory

orchestra/models.py
```diff
@@ -14,6 +12,56 @@
...
+        slug (str):
+            Unique identifier for the workflow version.
+        name (str):
+            Human-readable name for the workflow version.
+        description (str):
+            A longer description of the workflow version.
+        workflow (orchestra.models.Workflow):
+            The workflow that this is a version of.
+    """
+    slug = models.CharField(max_length=200)
```
marcua, line 58: should there be a separate code_directory for this one?  i'm personally alright with the argument that the slug should be tied to the directory unlike the workflow, but wanted to sanity check.

orchestra/models.py
```diff
@@ -28,16 +76,92 @@ class Certification(models.Model):
...
+        execution_function (str):
+            A JSON blob containing the path to and name of a python method that
+            will execute this step (only valid for machine steps).
+        required_certifications ([orchestra.models.Certification]):
+            The certifications a worker must have in order to work on this step
+            (only valid for human steps).
+        assignment_policy (str):
+            A JSON blob used to decide which worker to assign to this step
+            (only valid for human steps).
+        review_policy (str):
```
marcua, line 133: curiously is str or jsonfield a better description of this thing?

orchestra/models.py
```diff
@@ -28,16 +76,92 @@ class Certification(models.Model):
...
+        related_name='creation_dependents',
+        symmetrical=False)
+    submission_depends_on = models.ManyToManyField(
+        'self',
+        blank=True,
+        related_name='submission_dependents',
+        symmetrical=False)
+
+    # Machine step fields
+    is_machine = models.BooleanField()
```
marcua: an enum seems more fitting to me: https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/workflow.py#L143

orchestra/workflow/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class WorkflowError(Exception):
```
marcua: i think we have core.errors for stuff like this.  that said, i'd entertain us distributing errors to their "home module" in the future.

orchestra/workflow/directory.py
```diff
@@ -0,0 +1,36 @@
...
+        version_files = os.listdir(version_directory)
+        if 'version.json' not in version_files:
+            continue  # Subdirectory wasn't a workflow version.
+
+        with open(os.path.join(version_directory, 'version.json'), 'r') as f:
+            parsed['versions'].append(json.load(f))
+
+    # Complain if the workflow has no versions.
+    if len(parsed['versions']) == 0:
+        raise WorkflowError('Workflow directory {} does not contain any '
```
marcua, line 38: is this a log.error, or an exception.  trying to think of a time when you're just starting a workflow---would you ever not want the whole thing to crash?

orchestra/workflow/manifest.py
marcua: i would like to live in the future

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+            },
+            {
+                'name': certification_data['name'],
+                'description': certification_data['description'],
+            }
+        )
+
+    # Create the certification dependencies once all certs are in the db
+    for certification_data in workflow_data['certifications']:
+        cert = Certification.objects.get(
```
marcua: style: spell out variable names

orchestra/workflow/load.py
marcua, line 40: i don't think we want to do this, but do we want to delete objects that used to be in the workflow but aren't anymore?  this would break old versions, for example

orchestra/workflow/load.py
marcua: having a hard time visualizing this. as in, if the certification list expands but someone is already certified with it and shouldn't be?

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+from orchestra.models import Certification
+from orchestra.models import Workflow
+from orchestra.models import WorkflowStep
+from orchestra.models import WorkflowVersion
+from orchestra.workflow.directory import parse_workflow_directory
+from orchestra.workflow.exceptions import WorkflowError
+
+
+@transaction.atomic
+def load_workflow(workflow_directory, versions=None, force=False):
```
marcua: having read this, i feel like `versions` is too complicated.  i think that either you should pas in a single version and require the user to run this multiple times, or load all versions of a workflow.  thoughts?

orchestra/workflow/load.py
marcua, line 89: maybe just add a comment explaining the transactional safety part

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+        }
+    )
+
+    if not version_created:
+        if not force:
+            raise WorkflowError('Version {} already exists'
+                                .format(version_data['slug']))
+
+        # Check that the versions are safe to merge
+        new_step_slugs = set(s['slug'] for s in version_data['steps'])
```
marcua: style: s->step

orchestra/workflow/load.py
marcua, line 103: inverted example: if there are no projects for that workflow version, should we give them a free pass?  i'd opt for no, since a developer can always just delete those objects themselves

orchestra/models.py
```diff
@@ -28,16 +76,92 @@ class Certification(models.Model):
...
 
     def __str__(self):
         return '{}'.format(self.slug)
 
+    class Meta:
+        app_label = 'orchestra'
+        unique_together = (('workflow', 'slug'),)
+
+
+class WorkflowStep(models.Model):
```
marcua: Step or WorkflowStep?  Task is just a Task :)

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+                'description': step_data['description'],
+                'is_machine': step_data.get('is_machine', False),
+                'execution_function': step_data.get('execution_function', {}),
+                'assignment_policy': step_data.get('assignment_policy', {}),
+                'review_policy': step_data.get('review_policy', {}),
+                'user_interface': step_data.get('user_interface', {}),
+            }
+        )
+
+        # TODO(dhaas): should we prevent updates to these?
```
marcua: i could imagine over the course of developing a project, certification dependencies could be added or removed, and we don't want to uncertify workers automatically just because of that.

also: i'd be OK with us completely removing required_certifications, because we're just not there yet (TM) in terms of having dependency trees.  that said, i'm also OK just leaving things and not trying to be too prescriptive just yet

orchestra/workflow/load.py
marcua: query ahead of time and keep a dict of step->set(dependencies), and see if set=set?

marcua: also i agree that we want to prevent these and force a new version

orchestra/workflow/load.py
marcua: one more argument why we don't want to prevent too much: different certifications already mean different things across versions, so you're not preventing the workflow developer from much by trying to automatically determine if a certification was accidentally applied to a worker

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Set up step dependencies once the steps objects are in the DB.
+    # TODO(dhaas): should we prevent updates to these?
+    for step_data in version_data['steps']:
+        step = WorkflowStep.objects.get(
+            slug=step_data['slug'],
+            workflow_version=version
+        )
+        creation_depends_on_slugs = step_data.get('creation_depends_on', [])
+        creation_depends_on = WorkflowStep.objects.filter(
```
marcua: why does this not require a `list(...)` but line 126 does?

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+            workflow_version=version,
+            slug__in=creation_depends_on_slugs
+        )
+        step.creation_depends_on = creation_depends_on
+        submission_depends_on_slugs = step_data.get('submission_depends_on',
+                                                    [])
+        submission_depends_on = WorkflowStep.objects.filter(
```
marcua: same as comment above

orchestra/workflow/load.py
```diff
@@ -0,0 +1,161 @@
...
+            workflow_version=version,
+            slug__in=submission_depends_on_slugs
+        )
+        step.submission_depends_on = submission_depends_on
+
+
+def create_or_update(model, create_keys, update_keys):
```
marcua: seems like a common bit of functionality that we should surface, perhaps in `https://github.com/unlimitedlabs/orchestra/tree/master/orchestra/utils`

`utils.django`?

orchestra/management/commands/loadworkflow.py
thisisdhaas: whoops, copied that from the internets, will fix.

orchestra/models.py
thisisdhaas, line 34: Can you think of a scenario where we regret it? It would have to two different workflows needing to share the same code directory. I can't imagine this ever being a good idea, if we believe that workflows and their versions should be shared as Django apps.

orchestra/workflow/load.py
thisisdhaas: when you put it like that, I noticed this: https://docs.djangoproject.com/en/1.8/ref/models/querysets/#update-or-create

orchestra/management/commands/loadworkflow.py
thisisdhaas, line 13: Currently the semantics are "if it is provided, load the version. If not, load all versions." Are you saying it should be "a single version must be provided and that version will be loaded"?

orchestra/management/commands/loadworkflow.py
thisisdhaas: if it doesn't, a `WorkflowError` will be raised inside of `load_workflow`. Checking here would require duplicating some logic, I think.

Maybe the check for the workflow dir not existing (L. 26) should be moved inside `load_workflow` for consistency?

orchestra/migrations/0018_auto_20151007_1451.py
thisisdhaas, line 1: Added a comment.

orchestra/migrations/0018_auto_20151007_1451.py
thisisdhaas: Changed.

orchestra/migrations/0018_auto_20151007_1451.py
thisisdhaas: the only keys I'm aware of are to the Worker and to the Certification. Both are ForeignKeys, so there should be no cascades. 

The delete query django issues (against postgres) looks like:

```
DELETE FROM "orchestra_workercertification"
WHERE "orchestra_workercertification"."id" IN 
    (SELECT U0."id" AS Col1
     FROM "orchestra_workercertification" U0 
         INNER JOIN "orchestra_certification" U1 ON ( U0."certification_id" = U1."id" )
     WHERE U1."workflow_id" IS NULL)
```

which looks right to me. Agreed that this is scary :ghost:

orchestra/migrations/0018_auto_20151007_1451.py
thisisdhaas, line 79: ALL OF THE DATABASES. Except I don't think Sqlite even has an ALTER COLUMN command.

orchestra/models.py
thisisdhaas: Donesky

orchestra/models.py
thisisdhaas: implementado

orchestra/models.py
thisisdhaas, line 58: Right now there is no assumption made on the slug. The semantics are, 'A version is any direct subdirectory of a workflow's code_directory containing a valid `version.json`'. This lets us just search all subdirectories of the workflow for `version.json` files, and discover each version's slug. I'd be fine adding a code_directory for the version, but not all code for the version is required to be in the directory (e.g., static assets shared across versions), so it feels a bit odd.

orchestra/models.py
thisisdhaas: My preference is for being explicit--if I had my way, Task would be renamed :smile: 

But I don't feel strongly about it if you want to change it.

orchestra/models.py
thisisdhaas, line 133: I followed the existing convention I saw in other docstrings. jsonfield is probably better, unless it should be a dict (what you actually pass in to the field).

Happy to add an issue to clean it up everywhere if we decide on the right type.

orchestra/models.py
thisisdhaas: Hm, not sure I agree. Enums make code much more verbose (have to pass around `Step.WorkerType.HUMAN`s all over the place), and are unintuitive to serialize through JSON (i.e., our version.json manifests would have to either use the integer or some string name).

Can you think of a situation in the near-to-medium future where we have other types of workers? Right now, this field is mostly just sugar for detecting when a step has an execution_function vs. an interface.

orchestra/workflow/directory.py
thisisdhaas, line 38: I guess if you just wanted to test that your `workflow.json` parses? Since this code probably shouldn't be called except via `load_workflow()` though, I can't think of a good reason not to say that a versionless workflow is an invalid workflow.

orchestra/workflow/exceptions.py
thisisdhaas: I'll move this to core.errors for consistency now, and we can discuss later--I don't feel strongly one way or the other.

orchestra/workflow/load.py
thisisdhaas: The `loadworkflow.py` script only passes in one, so I kind of agree. But the only complexity avoided by switching to a single version argument is the set logic in lines 59-69. Most of the complexity comes from the fact that when running the function (with any number of versions), the workflow might already be in the DB even if a specific version is not.

orchestra/workflow/load.py
thisisdhaas, line 40: Yeah that seems bad. Certifications need to be append-only. The updates should be safe here since we're just changing name/description.

orchestra/workflow/load.py
thisisdhaas: certification is soooooolonggggg. I did it, even though it created several awkward line breaks.

orchestra/workflow/load.py
thisisdhaas: Yeah, but this is about dependencies. As in, I have Cert C, which has no required_certs. But someone reloads the workflow (or loads a new version), and the manifest now says that Cert C requires Cert B, which I don't have. Now I have an 'illegal' certification.

Should we: 
a) allow this situation, because it seems rare (and we want to let dependencies evolve over time).
b) throw an exception and refuse to change the certification dependencies.

orchestra/workflow/load.py
thisisdhaas, line 89: Added.

orchestra/workflow/load.py
thisisdhaas: Styled.

orchestra/workflow/load.py
thisisdhaas, line 103: Yeah, I think it's safer to ask the dev to go into the console and delete the WorkflowVersion in this case.

orchestra/workflow/load.py
thisisdhaas: To be clear, this comment is about preventing updates to a single workflow version step's set of required certifications, not any dependency trees of certifications. Do your comments still apply?

orchestra/workflow/load.py
thisisdhaas: because my excellent test suite would have caught this... IN THE FUTURE.

orchestra/workflow/load.py
thisisdhaas: Ok, will hwhip something up.

orchestra/workflow/load.py
thisisdhaas: same as comment above.

orchestra/workflow/load.py
thisisdhaas: Discussed in person. We'd like a workflow's certifications to evolve over time/versions, so we aren't going to restrict updates here. Users will have to manually add/remove certifications from workers appropriately.

orchestra/workflow/load.py
thisisdhaas: Re: conversation in person. We're going to allow certifications to evolve over time, and it's the responsibility of the user to make sure workers who shouldn't be certified have certifications removed.

Left a comment to that effect.

orchestra/models.py
```diff
@@ -1,19 +1,74 @@
-from datetime import datetime
 from django.contrib.auth.models import User
 from django.db import models
```
marcua, line 2: re: `I followed the existing convention I saw in other docstrings. jsonfield is probably better, unless it should be a dict (what you actually pass in to the field).

Happy to add an issue to clean it up everywhere if we decide on the right type.`

I think it should either be `dict` or `JSONField`.  Add an issue I guess?  I also feel like if you agree, we should set all of the stuff in this PR to be correct by our new standard.

orchestra/workflow/load.py
marcua: if i was folksy, i'd have an expression for this moment.  i really wish i was folksy

marcua:
Commented on https://github.com/unlimitedlabs/orchestra/commit/133841a1611316de0ce8901f12a0c0cc750b41b7

marcua:
Really minor comments on: https://github.com/unlimitedlabs/orchestra/commit/a1dc240633c87f7eff46c0bba3ecac22baf44ad6

marcua:
Small comments on https://github.com/unlimitedlabs/orchestra/commit/d1d50b8e50d73a24cca32f5094636dde3d916314

marcua:
Super minor comments on https://github.com/unlimitedlabs/orchestra/commit/d3c16891cfa2edb3986b3737decdcc0af7815ad7

marcua:
lgtm :shipit:

PR #59, opened 2015-10-19
Docs bugfixes, move to a separate requirements.txt file for read-the-docs.

thisisdhaas:
So it turns out that readthedocs doesn't actually need any of the requirements in `requirements.txt` currently to build the docs. This PR creates a dummy file that we can add to in `docs_requirements.txt`. In conjunction with changing the RTD settings, this will allow the docs to build without installing anything in `requirements.txt`.

As a result, no previous versions of our docs will build anymore. That should be fine, however--we'll be missing docs for 0.1.2 - 0.1.5 (which we can't build anyway) and things will return to normal when we release 0.1.6.

### DIFF
--- a/docs/source/core.rst
+++ b/docs/source/core.rst
@@ -1,23 +1,26 @@
 Core Reference
-=============
+==============
 **Core reference still in progress.**
 
 .. contents::
     :depth: 2
 
 .. workflow:
+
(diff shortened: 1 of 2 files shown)


marcua:
lgtm

PR #61, opened 2015-10-22
Add defaults to model JSONFields

jrbotros:
Adds a default empty object to all JSONFields in our models—otherwise, the default is an empty string, breaking some functionality.


marcua:
lgtm

PR #66, opened 2015-10-28
Fix project details screen and migrate project start_datetime

jrbotros:
When using `default=timezone.now` in a `DateTimeField`, Django provides a custom time-selector widget in the admin; however, when the model is saved, sub-second resolution is lost, which was causing a parsing error in the project details page. We've decided to leave the time editable in admin at the risk of losing sub-second accuracy in our data, but I've fixed the project details page bug and migrated `start_datetime` for projects to be consistent with the other models.

### DIFF
--- a/orchestra/admin_views.py
+++ b/orchestra/admin_views.py
@@ -1,8 +1,8 @@
-from datetime import datetime
 from django.contrib.admin.views.decorators import staff_member_required
 from django.shortcuts import render
 from django.utils.encoding import force_text
 from django.utils.text import capfirst
+from django.utils.dateparse import parse_datetime
 from orchestra.models import Project
 from orchestra.orchestra_api import get_project_information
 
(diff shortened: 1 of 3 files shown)


orchestra/admin_views.py
```diff
@@ -28,8 +28,7 @@ def project_details(request, project_id):
 
 
 def _prettify_project_details(project, tasks_data, steps):
-    project['start_datetime'] = datetime.strptime(project['start_datetime'],
-                                                  '%Y-%m-%dT%H:%M:%S.%fZ')
+    project['start_datetime'] = parse_datetime(project['start_datetime'])
```
thisisdhaas, line 31: Just confirming that parse_datetime will make this timezone-aware. Is there an option to tell it to use UTC?

thisisdhaas:
LGTM if you answer my question.

orchestra/admin_views.py
jrbotros, line 31: It's using the set timezone for the project—`start_datetime` gets stored as timezone-aware, and `parse_datetime` will read it out in the same way.

PR #69, opened 2015-11-03
Updates to the latency analytics api, a completed projects filter, and two convenience aggregates

marcua:


orchestra/analytics/latency.py
```diff
@@ -51,6 +53,8 @@ def work_time_df(projects, human_only=True, complete_tasks_only=True):
             human_only, complete_tasks_only)
         for project in projects))
     df = DataFrame(time_data)
+    df['start_datetime'] = df['start_datetime'].astype('datetime64[ns]')
```
thisisdhaas, line 60: What does this do? Why were things working without it? Maybe an explanatory comment? Why is this done after constructing the dataframe?

orchestra/analytics/latency.py
```diff
@@ -51,6 +53,8 @@ def work_time_df(projects, human_only=True, complete_tasks_only=True):
...
+    df['end_datetime'] = df.start_datetime + df.calendar_time
```
thisisdhaas: L128 seems to use end_datetime during the construction of the dataframe—why not add this column there?

orchestra/analytics/latency.py
```diff
@@ -128,3 +132,55 @@ def project_time_row_generator(project_information,
...
+        df (pandas.DataFrame):
+            A DataFrame with a schema described in `work_time_df`.
+        dimensions (list):
+            A list of column names to group on.
+
+    Returns:
+        df (pandas.DataFrame):
+            A DataFrame with aggregate information.
+    """
+    return df.groupby(dimensions, as_index=False).agg({'work_time': np.sum})
```
thisisdhaas, line 160: - Is it Orchestra's job to provide analytics convenience methods?
- This is a really thin wrapper around pandas—how useful is it?
- what does `as_index` do?

orchestra/analytics/latency.py
```diff
@@ -128,3 +132,55 @@ def project_time_row_generator(project_information,
...
+            A DataFrame with aggregate information.
+    """
+    # When reporting the project time, we want to avoid double-counting two
+    # tasks that are running in parallel, so we calculate the difference
+    # between the farthest-apart start and end times for iterations.  For more
+    # granular groupings like assignments, because iterations are
+    # interspersed, computing the end time minus the start time would double-
+    # count calendar time, so we want to sum up their computed
+    # calendar_datetimes.
+    if dimensions == ['project_id']:
```
thisisdhaas: what if `project_id` is one of several aggregating dimensions?

orchestra/analytics/latency.py
```diff
@@ -128,3 +132,55 @@ def project_time_row_generator(project_information,
...
+        df (pandas.DataFrame):
+            A DataFrame with aggregate information.
+    """
+    # When reporting the project time, we want to avoid double-counting two
+    # tasks that are running in parallel, so we calculate the difference
+    # between the farthest-apart start and end times for iterations.  For more
+    # granular groupings like assignments, because iterations are
+    # interspersed, computing the end time minus the start time would double-
+    # count calendar time, so we want to sum up their computed
+    # calendar_datetimes.
```
thisisdhaas: This comment is very hard to read and needs to be simplified/clarified.

After reading the code, the first sentence of the comment seems to capture all of important information?

orchestra/analytics/latency.py
```diff
@@ -128,3 +132,55 @@ def project_time_row_generator(project_information,
...
+    if dimensions == ['project_id']:
+        aggregated = df.groupby(dimensions, as_index=False).agg({
+            'start_datetime': np.min,
+            'end_datetime': np.max})
+        aggregated['calendar_time'] = (
+            aggregated['end_datetime'] - aggregated['start_datetime'])
+        aggregated.drop('start_datetime', axis=1, inplace=True)
+        aggregated.drop('end_datetime', axis=1, inplace=True)
+        return aggregated
```
thisisdhaas, line 192: sanity check: does this do the right thing with iteration 0 (this applies to the non-project_id case below as well).

orchestra/analytics/latency.py
```diff
@@ -128,3 +132,55 @@ def project_time_row_generator(project_information,
...
+    else:
+        return df.groupby(dimensions, as_index=False).agg(
+            {'calendar_time': np.sum})
```
thisisdhaas, line 200: This function and the previous are called `_aggregate`, but they only compute sums. Should we pass in the aggregate function, or be more explicit with the name?

orchestra/utils/project_properties.py
```diff
@@ -0,0 +1,22 @@
...
+
+def completed_projects(projects):
+    """
+    Filters `projects` queryset to completed ones.
+
+    TODO(marcua): if we use this for business logic, we should also
+    ensure that the number of tasks equals the number of steps in
+    the workflow, which will be easier after @thisisdhaas implements
+    workflows-near-the-db.
+    """
```
thisisdhaas, line 14: Also add a note that it would be cleaner to just add a Complete state to Projects.

orchestra/analytics/latency.py
marcua, line 60: Nice!  Added a comment.

orchestra/analytics/latency.py
marcua: Good catch, moved it

orchestra/analytics/latency.py
marcua, line 160: To the first two questions: absolutely.  work_time_aggregate's implementation is simple, but calendar_time_aggregate's implementation is not.  Providing the functions encapsulates the challenging logic so that people who want to work with the data have a better abstraction.

To the third question: Pandas `groupby` is more like Pig's grouping operator than SQL's.  It creates indexed/hierarchically clustered keys that map on to the remaining ungrouped columns. `as_index=False` flattens things out so the result is a lot more like a relation/SQL group by.  Is this worth a comment?  I'd have to add it to both places where I group, but it feels a bit idiosyncratic to Pandas.

orchestra/analytics/latency.py
marcua: I think we're fine in that case.  That's the `else`, which is explained in the second sentence of this paragraph-comment :).

orchestra/analytics/latency.py
marcua: I agree that this comment is long and dense.  I disagree in that the second sentence is critical to understanding the `else:` clause.  Would you recommend I put that second sentence near the `else:`?  Happy to chat offline real quick

orchestra/analytics/latency.py
marcua, line 192: I think so, but vet this:
- For automatically assigned tasks, the calendar time is nearly 0, so the start time of iterations 0 and 1 are nearly identical.
- For non-automatically assigned tasks, the calendar time is non-zero, depending on when an expert picks up a task.    The calendar time of the task/project, however, should include this non-zero time, which is captured by its earlier `start_datetime`.

In the `else:` clause below, we similarly want to include the calendar time that an expert didn't pick the time up for when considering the calendar time of a project/task/etc.

orchestra/analytics/latency.py
marcua: I'm breaking the comment into smaller ones, and realizing that I'm not sure we have a good semantic definition of calendar time for things like assignments.  I'll reach out offline to figure it out

orchestra/analytics/latency.py
marcua, line 200: Good call---changed to _sums

orchestra/utils/project_properties.py
marcua, line 14: Good idea and done :)

orchestra/analytics/latency.py
marcua: Updated: I broke the comment down and tried to clarify.  How's that work?

orchestra/analytics/latency.py
jrbotros, line 60: This comment still seems confusing to me—if pandas treats non-primitives as objects and we're storing the datetime objects in pandas originally, why the need to cast?

orchestra/analytics/latency.py
jrbotros: :white_check_mark:

orchestra/analytics/latency.py
jrbotros, line 160: Sounds good to me on the first two counts. I think `as_index=False` is a pretty common pandas paradigm, so I don't think it warrants a comment.

orchestra/analytics/latency.py
jrbotros: Looking better now!

orchestra/analytics/latency.py
jrbotros, line 192: Sounds good to me.

jrbotros:
lgtm

marcua:
Addressed comments here: https://github.com/unlimitedlabs/orchestra/commit/75406a17c3a2d5e0dcf694fcee7e986b1cfcf12f

PR #71, opened 2015-11-11
Slack naming

marcua:


orchestra/slack.py
```diff
@@ -50,9 +50,25 @@ def create_project_slack_group(project):
     return project.slack_group_id
 
 
+def _random_string():
+    return ''.join(
+        random.choice('abcdefghijklmnopqrstuvwxyz1234567890')
```
thisisdhaas: I think it's preferred to do `import string` and `random.choice(string.ascii_lowercase + string.digits)`

orchestra/tests/test_notifications.py
```diff
@@ -219,3 +220,49 @@ def _validate_slack_messages(message_stub):
...
+        self.assertTrue(group_name.startswith('sallys-plumbingo'))
+        self.assertEqual(len(group_name), 21)
+
+        # Test that we create unique slack IDs if there are conflicts by
+        # mocking the randomization logic to return deterministic results and
+        # ensure that we don't repeat project IDs.
+        with patch('orchestra.slack._random_string', new=fake_random_string):
+            for short_description, group_id in (
+                    ('ketchup #3 Sales --:/ kitchen supplies',
+                     'ketchup-3-sales-2'),
```
thisisdhaas, line 261: Shouldn't this start at `-1`?

orchestra/tests/test_notifications.py
```diff
@@ -219,3 +220,49 @@ def _validate_slack_messages(message_stub):
...
+                    ('!!bongo bash*&#$(*$7235818923701842312',
+                     'bongo-bash723581-5'),
```
thisisdhaas, line 266: and why does this skip to 5?

orchestra/slack.py
```diff
@@ -50,9 +50,25 @@ def create_project_slack_group(project):
...
-    return base64.b64encode(uuid1().bytes)
+    name = None
+    # The human-readable portion of the name (16 characters) involves
+    # slugifying the project short description.
+    descriptor = slugify(project.short_description)[:16].strip('-')
+    while True:
+        # Add 4 characters of randomness (~1.68 million permutations).
+        name = '{}-{}'.format(descriptor, _random_string())
+        if not Project.objects.filter(slack_group_id=name).exists():
+            break
```
thisisdhaas, line 75: Maybe we need to actually check for duplicates with the slack APIs? As in, what if we have groups left over from testing/before the last DB reset or created for other reasons?

And if it turns out not to be unique, what's the failure mode? Does it error? post messages to the existing group? Overwrite the existing group?

orchestra/slack.py
marcua: done

orchestra/tests/test_notifications.py
marcua, line 261: see if it all makes sense in the new code

orchestra/tests/test_notifications.py
marcua, line 266: see if it all makes sense in the new code

orchestra/slack.py
marcua, line 75: done---we were being way too loose with group ids vs group names, so this resulted in a bunch of modifications to testing code.

orchestra/slack.py
```diff
@@ -50,9 +50,27 @@ def create_project_slack_group(project):
...
+
+    Slack group names are capped at 21 characters in length.
     """
-    return base64.b64encode(uuid1().bytes)
+    name = None
+    # The human-readable portion of the name (16 characters) involves
+    # slugifying the project short description.
+    descriptor = slugify(project.short_description)[:16].strip('-')
+    slack = SlackService(settings.SLACK_EXPERTS_API_KEY)
+    groups = {group['name'] for group in slack.groups.list().body['groups']}
```
thisisdhaas, line 70: sanity check--this is wrapped in a @run-if somewhere?

orchestra/tests/test_notifications.py
thisisdhaas, line 261: oic

orchestra/tests/test_notifications.py
thisisdhaas, line 266: maybe this needs explicit comments of the form "-5 will be used, because the mock function's counter is at 3 and -3 and -4 already exist" -- I had to stare at this for a while before I got it.

thisisdhaas:
LGTM after adding a comment.

orchestra/slack.py
marcua, line 70: good question!  it's only called by `create_project_slack_group`, which is wrapped

PR #72, opened 2015-11-13
Add favicon to Orchestra

jrbotros:
orchestra


marcua:
lgtm but check FF and phone

PR #73, opened 2015-11-17
Handles notifications for legacy projects with no slack channel.

thisisdhaas:


jrbotros:
lgtm

orchestra/utils/notifications.py
```diff
@@ -84,7 +84,8 @@ def notify_status_change(task, previous_status=None):
         }
 
     _notify_internal_slack_status_change(task, current_worker)
-    _notify_experts_slack_status_change(task, current_worker)
+    if task.project.slack_group_id:
```
jrbotros, line 87: Will this error out if the worker doesn't have a slack handle?

jrbotros:
One comment, then lgtm

orchestra/utils/notifications.py
thisisdhaas, line 87: this line won't--it doesn't use the worker at all. 

Line 132, below, makes sure that there is no @-mention if the worker doesn't have a handle.

jrbotros, line 87: Sweet, sounds good.

PR #74, opened 2015-11-24
Implement view for visualizing and administrating projects

jrbotros:
This view provides a visualization for project data and should allow users to circumvent the project and task admin sites for most all common actions, providing functionality for:
- reverting a task to a previous state
- reassigning any assignment for a given task
- adding a new assignment for `AWAITING_PROCESSING` or `PENDING_REVIEW` tasks
- skipping a task and marking it as complete (e.g., a machine task that hangs and prevents subsequent task creation)
- adding and removing slack users from the project group
- creating subsequent tasks for the project
- viewing data and linking to the Django admin at all levels of granularity

To view the project management screen for any project, you'll need to be logged into a user who's been added to a `project_admins` Django group.


orchestra/analytics/latency.py
```diff
@@ -128,7 +135,14 @@ def project_time_row_generator(project_information,
                     assignment['worker'],
                     timedelta(seconds=snapshot['work_time_seconds']),
                     parse(snapshot['datetime'])))
-
+            if in_progress_iterations and assignment['status'] == 'Processing':
```
marcua: note to remind us to talk about iteration 0, etc.

orchestra/analytics/latency.py
```diff
@@ -128,7 +135,14 @@ def project_time_row_generator(project_information,
...
+                iterations.append(Iteration(
```
marcua: it looks like there are three ways to append to iterations atm---can you describe all three cases?

orchestra/api_urls.py
```diff
@@ -1,7 +1,15 @@
 from django.conf.urls import patterns
 from django.conf.urls import url
+from orchestra.views import assign_task_api
```
marcua: these functions should likely not all end in `_api`.  instead, consider putting them in their own set of project management views (e.g., `project_management_api_views.py`)

alternatively: move all of the `/orchestra/api/interface/*` views in to an interface_api_views.py

orchestra/static/orchestra/common/css/common.scss
```diff
@@ -28,7 +28,9 @@ $panel-button-margin: 1.5rem;
 }
 
 .disable-select {
```
marcua, line 30: reminder to see if it's easy to fix the UI on safari

orchestra/static/orchestra/common/js/orchestra_api.js
```diff
@@ -0,0 +1,62 @@
...
+      return $http.post('/orchestra/api/interface/reassign/', {
+        'worker_username': worker_username,
+        'assignment_id': assignment.id
+      })
+    }
+
+    actionService.revert = function(taskId, datetime, fake) {
+      return $http.post('/orchestra/api/interface/revert_task/', {
+        'taskId': taskId,
+        'revert_datetime': datetime.getTime() - 1000,
```
marcua: what's this `- 1000` stuff?

orchestra/static/orchestra/common/js/orchestra_api.js
```diff
@@ -0,0 +1,62 @@
+var serviceModule =  angular.module('orchestra.common.services');
+
+serviceModule.factory('orchestraApi', ['$http', function($http) {
+  var actionService = new function() {
+    var actionService = this;
```
marcua: sanity check: do the other `/orchestra/api/interface/` views (the ones that existed before this PR) also get served from this service?

orchestra/static/orchestra/common/js/orchestra_api.js
```diff
@@ -0,0 +1,62 @@
+var serviceModule =  angular.module('orchestra.common.services');
+
```
marcua, line 2: high-level: make the url_endpoints and `methodNames` be the same (e.g., `edit_slack_membership` vs `editSlackGroupMembership` or `complete_and_skip_task` vs `skipTask`)

orchestra/static/orchestra/common/js/orchestra_api.js
```diff
@@ -0,0 +1,62 @@
...
+serviceModule.factory('orchestraApi', ['$http', function($http) {
+  var actionService = new function() {
+    var actionService = this;
+
+    actionService.projectManagementData = function(projectId) {
+      return $http.post('/orchestra/api/interface/project_management_information/', {
```
marcua: minor nitpick: consider avoiding repeating `/orchestra/api/interface/` in case we change that URL---have it in one variable?

OK if you don't change :)

orchestra/static/orchestra/project/css/project.scss
```diff
@@ -0,0 +1,163 @@
...
+    margin: 0 10px;
+  }
+
+  .axis-wrapper {
+    position: relative;
+    display: inline-block;
+  }
+
+  .axis {
+    shape-rendering: crispEdges;
```
marcua: this and line 70 are highlighted.  any idea why?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
+(function () {
```
marcua: call it project_management, not project

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+  // 'use strict';
+
+  angular
+    .module('orchestra.project.controllers')
```
marcua: project_management

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+    .controller('ProjectController', ProjectController);
+
+  ProjectController.$inject = ['$location', '$scope', '$routeParams', '$http', '$sce',
```
marcua: ProjectManagementController

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                            '$compile', '$modal', '$timeout', 'dataService', 'orchestraApi'];
+
+  function ProjectController($location, $scope, $routeParams, $http, $sce,
+                          $compile, $modal, $timeout, dataService, orchestraApi) {
+    // var vm = this;
```
marcua: ?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+    vm = $scope;
```
marcua: var

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+    $scope.activate = function() {
+      // Prevent initial animation of vis load
+      d3.select('.project-management').style('opacity', 0);
```
marcua: see my next comment about this, to see why i suggest this, but: encapsulate all d3 calls inside the *Vis functions.  that will allow you to have an angular controller/file without any implementation details, and you'll be able to put the vis in a different non-angular module/file

this would mean embedding lines 17 and 19-22 inside the projectVis constructor

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+      $scope.vis = new projectVis($routeParams.projectId, '.project-management .svg-wrapper');
+      $scope.vis.update();
+      d3.select('.project-management').transition()
+        .delay(500)
+        .style('opacity', 1);
+    };
+    $scope.activate();
+
+    function projectVis(projectId, parentSelector) {
```
marcua: Throw all of the *Vis functions inside their own .js file.  if you want, you can go so far as to make one file per *Vis function, but I don't think you need to go that far.

This way, you'll separate the d3 stuff from the angular stuff.

marcua: continuing the nitpick: projectManagementVis

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+      }
+
+      vis.relativeTime = true;
+      vis.scaleWidth = 1350;
+
+      vis.crosshair = new crosshair();
+      vis.actions = new actions();
+      vis.utils = new utils();
+      vis.tasks = new tasksVis();
+      vis.assignments = new assignmentsVis();
```
marcua: style: you're using `new` and `class`-like patterns for creating each of these objects.  But classes are typically used for encapsulating state.  A lot of the code for things like `assignments` and other variables here is just a nice modularization of your logic.  Since the modules don't maintain/encapsulate internal state, it would be better (stylistically) to avoid using `new`, and instead follow the module/singleton pattern: http://stackoverflow.com/a/1479341

it won't involve crazy changes---basically, rather than assigning functions on `this` in each module, you return an object/dict with all of the public methods you want to export.

edit: you could also make the vis stuff its own service if you want to continue using angular to build your services.  it would still separate the d3/vis from the controller.

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          'width': vis.utils.getSvgWidth(),
+          'height': vis.options.scaleHeight,
+        })
+        .style('margin-left', vis.options.marginLeft);
+
+      axisWrapper.append('span')
+        .attr('class', 'x label')
+
+      vis.draw = function() {
+        // vis.drawSvg();
```
marcua: haaay

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          .then(function(response) {
+            vis.data = response.data;
+            vis.prepareData();
+            dataService.data(vis.data);
+            vis.draw();
+          });
+      }
+
+      vis.prepareData = function() {
+        _.each(vis.data.tasks, function(task) {
```
marcua: can we move all `each` instances over to `array.forEach`, e.g., `task.assignments.forEach`?

i don't feel strongly abut this since you use `_.map` and `_.max` elsewhere.  perhaps only make this change if you can completely eliminate underscore from this implementation.  if you can't, then keep `_.each` for consistency.

high level: i generally like to push against adding new library dependencies.  here we added d3, and we should be really sure we also want to add underscore

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                             });
+        vis.timeSortedSlugs = _.map(timeSortedTasks, function(task) {return task.step_slug})
+      }
+
+      vis.setupAxis = function() {
+        var firstTask = vis.data.tasks[_.first(vis.timeSortedSlugs)];
+        minDatetime = firstTask.start_datetime.toDate();
+
+        var taskEndTimes = _.map(vis.timeSortedSlugs,
+                                 function(slug) {return dataService.taskEnd(dataService.taskFromKey(slug))})
```
marcua: style: for something this long, it no longer feels like it should be a single-liner---consider putting on multiple lines

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+
+        var timeSortedTasks = _.toArray(vis.data.tasks).sort(function(a, b) {
+                                return d3.ascending(a.start_datetime.toDate(), b.start_datetime.toDate());
+                             });
+        vis.timeSortedSlugs = _.map(timeSortedTasks, function(task) {return task.step_slug})
+      }
+
+      vis.setupAxis = function() {
+        var firstTask = vis.data.tasks[_.first(vis.timeSortedSlugs)];
+        minDatetime = firstTask.start_datetime.toDate();
```
marcua: is this guaranteed to be the min?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+        })
+
+        var timeSortedTasks = _.toArray(vis.data.tasks).sort(function(a, b) {
+                                return d3.ascending(a.start_datetime.toDate(), b.start_datetime.toDate());
+                             });
+        vis.timeSortedSlugs = _.map(timeSortedTasks, function(task) {return task.step_slug})
+      }
+
+      vis.setupAxis = function() {
+        var firstTask = vis.data.tasks[_.first(vis.timeSortedSlugs)];
```
marcua: is this somehow better than `vis.timeSortedSlugs[0]`?

marcua: if there are 0 slugs, do we get an exception/

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          .attr({
+            'width': vis.utils.getSvgWidth()
+          })
+      }
+
+      function tasksVis() {
+        var tasksVis = this;
+
+        tasksVis.draw = function() {
+          tasksVis.drawMeta();
```
marcua: i see a few of these `drawMeta` around, and don't know what they do.  can you change the name to something more informative or add some comments?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          .style('right', vis.utils.getSvgWidth() + 5 + 'px');
+
+        vis.parentContainer.select('.x.axis').transition().call(xAxis);
+        d3.select('.x.axis')
+          .attr({
+            'width': vis.utils.getSvgWidth()
+          })
+      }
+
+      function tasksVis() {
```
marcua: consider adding a short docstring to each module describing what it's responsible for

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                    $scope.cancel();
+                  });
+
+                modalInstance.result.finally(function() {
+                  vis.reverting = false;
+                  vis.crosshair.hide();
+                });
+              })
+
+              function humanizeAudit(audit) {
```
marcua: cute name

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                console.log(audit)
```
marcua: remove all `console.log`s throughout

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                  }
+                });
+                return humanAudit
+              }
+            })
+
+          revertGroups.transition().attr({
+            'transform': function(datetimeInfo) {
+              var taskStartDatetime = dataService.taskFromKey(datetimeInfo.taskKey).start_datetime;
+              return vis.utils.translateString(vis.timeScale(datetimeInfo.datetime) - vis.utils.getOffset(taskStartDatetime), 0);
```
marcua: this line is very long

https://www.youtube.com/watch?v=kkwiQmGWK4c

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                }
+                else {
+                  return vis.options.barHeight + 1;
+                }
+              },
+            })
+        }
+
+        tasksVis.expandDiagrams = function(taskObj) {
+          // var slug = taskObj.datum();
```
marcua: ohai

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                  return (task.assignments.length + 1) * vis.options.barHeight + 1;
+                }
+                else {
+                  return vis.options.barHeight + 1;
+                }
+              },
+            })
+        }
+
+        tasksVis.expandDiagrams = function(taskObj) {
```
marcua: what's this function do?  (add comment)

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+            return task.step_slug;
+          });
+
+          taskNames.selectAll('.step-status').text(function(slug) {
+            var task = dataService.taskFromKey(slug);
+            return task.status;
+          });
+        }
+
+        tasksVis.drawDiagrams = function() {
```
marcua: comment/docstring

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+                vis.crosshair.hide();
+              }
+            })
+            .on('click', function(datetimeInfo) {
+              if (vis.reverting) {
+                return;
+              }
+              vis.reverting = true;
+              var taskId = dataService.taskFromKey(datetimeInfo.taskKey).id;
+              orchestraApi.revert(taskId, datetimeInfo.datetime, 'fake')
```
marcua: i you split this out, you might implement it in the angular code and pass in a callback/function to the projectVis that facilitates this

marcua: 'fake' is a funny argument, versus passing in true/false for a variable called `fake`

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+
+          assignmentsVis.drawMeta();
+        }
+
+        assignmentsVis.drawMeta = function() {
+          // Draw divs on top
+          var assignmentsMeta = vis.parentContainer
+            .selectAll('.task-view')
+            .selectAll('.assignment-meta')
+            .data(function(taskKey) {return _.map(dataService.taskFromKey(taskKey).assignments, dataService.keyFromAssignment)});
```
marcua: long :)

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+            })
+            .on('mouseenter', function() {
+              d3.select(this).attr('stroke', 'yellow');
+            })
+            .on('mouseleave', function() {
+              d3.select(this).attr('stroke', 'none');
+            })
+            .on('click', function(iterationKey, i) {
+              var iteration = dataService.iterationFromKey(iterationKey);
+              var modalInstance = $modal.open({
```
marcua: similar story for modals as the one above

or have a modalservice...

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+            'x1': 0,
+            'x2': 0,
+            'y1': 0,
+          });
+
+        this.draw = function() {
+          this.svg
+            .attr({
+              'height': vis.utils.getSvgHeight(),
+              'width': 1 // vis.utils.getSvgWidth()
```
marcua: yaw

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+        this.show = function() {
+          this.draw();
+          this.svg.attr('opacity', 1);
+        }
+        this.hide = function() {
+          this.svg.attr('opacity', 0);
+        }
+        this.move = function(datetime) {
+          this.svg.style('left', (vis.options.marginLeft + vis.timeScale(datetime)) + 'px');
+          // this.line.attr('transform', vis.utils.translateString(vis.timeScale(datetime), 0))
```
marcua: haaay

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+
+        actions.createSubsequentTasks = function() {
+          orchestraApi.createSubsequentTasks(vis.projectId)
+            .then(function() {vis.update()},
+                  function() {alert('Could not create subsequent tasks.')});
+        }
+
+        actions.skipTask = function(task) {
+          if (!confirm('Are you sure you want to skip this task and mark it as ' +
+                      'complete? All sorts of crazy things might happen.')) {
```
marcua: replace last sentence with "This might leave the project in a corrupted/unrecoverable state."

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          this.move(vis.timeScale.domain()[0]);
+        }
+      }
+
+      function actions() {
+        var actions = this;
+
+        actions.createSubsequentTasks = function() {
+          orchestraApi.createSubsequentTasks(vis.projectId)
+            .then(function() {vis.update()},
```
marcua: high-level: these one-liner functions are very d3-esque, and not very nice more broadly.  i'd vote for this to not become our style, and for us to revert this style in this file except maybe with the exception of places where it's d3-esque.  agree?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+              alert('Error reassigning worker.')
+            })
+            .finally(function() {
+              inputEl.node().blur();
+              inputEl.node().value = assignment.worker;
+              assignment.reassigning = false
+            })
+        }
+
+        actions.showProjectData = function() {
```
marcua: sometimes the modal is in its own function, and sometimes it's baked into a larger function.  i like it when it's on its own like it is here

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
...
+          this.svg.style('left', (vis.options.marginLeft + vis.timeScale(datetime)) + 'px');
+          // this.line.attr('transform', vis.utils.translateString(vis.timeScale(datetime), 0))
+        }
+        this.reset = function() {
+          this.svg.attr('opacity', 0);
+          this.move(vis.timeScale.domain()[0]);
+        }
+      }
+
+      function actions() {
```
marcua: projectActions?

i say this because there are also taskActions

orchestra/static/orchestra/project/js/data_service.js
```diff
@@ -0,0 +1,90 @@
+var serviceModule =  angular.module('orchestra.project.services');
+
+serviceModule.factory('dataService', function() {
+  var dataService = new function() {
+    var that = this;
```
marcua: better name than `that`:P

orchestra/static/orchestra/project/js/data_service.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Iteration helpers
+    this.iterationFromKey = function(key) {
+      return this.assignmentFromKey(key.assignmentKey).iterations[key.iterationIndex];
+    }
+
+    this.keyFromIteration = function(iteration) {
+      return {
+        'assignmentKey': this.keyFromAssignment(iteration.assignment),
```
marcua: this, or that?  my guess is that `this` is wrong for this line

orchestra/static/orchestra/project/js/data_service.js
```diff
@@ -0,0 +1,90 @@
+var serviceModule =  angular.module('orchestra.project.services');
+
+serviceModule.factory('dataService', function() {
+  var dataService = new function() {
+    var that = this;
+    this._meta = {'tasks': {}};
```
marcua: feels funny to underscore this, since private/public in javascript is handled by only returning an object with the public methods.

marcua: there might be a better name for this than meta, since it's the actual task data, no?

orchestra/static/orchestra/project/js/controllers.js
```diff
@@ -0,0 +1,870 @@
+(function () {
+  // 'use strict';
+
+  angular
+    .module('orchestra.project.controllers')
+    .controller('ProjectController', ProjectController);
+
+  ProjectController.$inject = ['$location', '$scope', '$routeParams', '$http', '$sce',
+                            '$compile', '$modal', '$timeout', 'dataService', 'orchestraApi'];
```
marcua: is angular style that services are prepended with $, and if so, should you prepend the service and orchestra api?

orchestra/static/orchestra/project/js/data_service.js
```diff
@@ -0,0 +1,90 @@
...
+
+    this.data = function(data) {
```
marcua: `this.setData` (or maybe reset rather than set)

orchestra/tests/test_task_lifecycle.py
```diff
@@ -17,16 +19,19 @@
 from orchestra.tests.helpers.fixtures import setup_models
 from orchestra.tests.helpers.fixtures import StepFactory
 from orchestra.utils.assignment_snapshots import empty_snapshots
-from orchestra.utils.task_lifecycle import _worker_certified_for_task
 from orchestra.utils.task_lifecycle import _check_worker_allowed_new_assignment
+# from orchestra.utils.task_lifecycle import _get_last_active_assignment
```
marcua: heya

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
         self.assertTrue(worker_has_reviewer_status(self.workers[6]))
```
marcua, line 171: high-level testing question: is 100% of the new Python you wrote covered?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
 
     def test_assign_task(self):
+        entry_task = Task.objects.create(
```
marcua: use TaskFactory?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+                invalid_status_task = assign_task(
+                    self.workers[0].id, invalid_status_task.id)
+
+        # Attempt to assign review task to worker already in review hierarchy
+        review_task = Task.objects.create(
+            project=self.projects['base_test_project'],
+            status=Task.Status.PENDING_REVIEW,
+            step=self.test_step)
+        test_data = {'test_assign': True}
+        TaskAssignment.objects.create(worker=self.workers[1],
```
marcua: factory

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
         self.assertTrue(worker_has_reviewer_status(self.workers[6]))
 
     def test_assign_task(self):
```
marcua, line 172: sanity checks:
- are you still covering all of the assignment functionality?
- there's a lot of code deleted in this test.  did we delete any semantically-relevant conditions that we want to put back in?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -494,18 +628,6 @@ def test_legal_get_next_task_status(self):
                                  TaskAssignment.SnapshotType.ACCEPT),
             Task.Status.COMPLETE)
 
-    def test_task_history_details(self):
```
marcua: why delete?

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+    task.save()
+    for assignment in task.assignments.all():
+        assignment.status = TaskAssignment.Status.SUBMITTED
+        assignment.save()
+    create_subsequent_tasks(task.project)
+    return task
+
+
+@transaction.atomic
+def revert_task_to_datetime(task_id, revert_datetime, fake=False):
```
marcua, line 274: note: I don't love the idea that we're reverting to a datetime rather than reverting to a particular assignment/iteration, since time is a super fuzzy and inaccurate thing in running systems.  We've got assignments/iteration counters to provide a semantic ordering over the iterations, and ideally we'd use that.

But I also realize that implementing `revert_task_to_iteration` might be harder to implement, and if it's significantly harder, I'm not suggesting you do it in the interest of getting something out there:).

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+        first_review_assignment = task.assignments.get(assignment_counter=2)
+        entry_assignment = reassign_assignment(
+            self.workers[3].id, entry_assignment.id)
+        self.assertEquals(entry_assignment.worker, self.workers[3])
+
+    def test_revert_task_to_datetime(self):
+        test_start = timezone.now()
+
+        def _sample_complete_task():
+            task = Task.objects.create(
```
marcua: factory

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+        # Reassign review assignment to another reviewer
+        first_review_assignment = task.assignments.get(assignment_counter=2)
+        entry_assignment = reassign_assignment(
+            self.workers[3].id, entry_assignment.id)
+        self.assertEquals(entry_assignment.worker, self.workers[3])
+
+    def test_revert_task_to_datetime(self):
+        test_start = timezone.now()
+
+        def _sample_complete_task():
```
marcua: this feels like it should live in helpers/fixtures.  I'd move it there

marcua: sample is an overloaded term---setup/create might be better here

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+                },
+                {
+                    'worker': workers['reviewer'],
+                    'type': TaskAssignment.SnapshotType.ACCEPT,
+                    'datetime': test_start + timedelta(hours=6),
+                },
+            ]
+
+            for i, assignment in enumerate(assignments):
+                TaskAssignment.objects.create(
```
marcua: factory

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+                    worker=assignment['worker'], task=task,
+                    status=TaskAssignment.Status.SUBMITTED,
+                    in_progress_task_data={},
+                    assignment_counter=i,
+                    snapshots={'snapshots': [], '__version__': 2},
```
marcua: use this function: https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/utils/assignment_snapshots.py#L14

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+                        snapshot['worker'].user.username: snapshot['type']
+                    },
+                })
+                assignment.save()
+
+            return task
+
+        def _test_reverted_task(datetime, status, num_assignments,
+                                num_snapshots_per_assignment):
+            task = _sample_complete_task()
```
marcua: is there any benefit to testing this on incomplete tasks?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+            revert_task_to_datetime(task.id, datetime)
+            task = Task.objects.get(id=task.id)
+            self.assertEquals(task.status, status)
+
+            assignments = assignment_history(task)
+            self.assertEquals(task.assignments.count(), num_assignments)
+            self.assertEquals(
+                len(num_snapshots_per_assignment), num_assignments)
+            for i, num_snapshots in enumerate(num_snapshots_per_assignment):
+                self.assertEquals(
```
marcua: also: create assignment snapshot data that's unique to each submission, and ensure that the snapshot's data is removed while the active data the worker sees is equal to the removed snapshot's data.

marcua: sanity check: did you also test by hand that this is the case (reverting to a state means the previous worker sees all of their work preserved?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -17,16 +19,19 @@
 from orchestra.tests.helpers.fixtures import setup_models
 from orchestra.tests.helpers.fixtures import StepFactory
 from orchestra.utils.assignment_snapshots import empty_snapshots
```
marcua, line 20: high-level: i'd feel more comfortable about these new tests if they tested at the views.py/request level rather than internal apis.  while we try to not include business logic in views.py, we inevitably fail, and ensuring we test this (plus error conditions) keeps us less brittle.

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+        with self.assertRaises(Task.DoesNotExist):
+            Task.objects.get(id=task.id)
+
+    def test_complete_and_skip_task(self):
+        for task in Task.objects.all():
+            task = complete_and_skip_task(task.id)
+            self.assertEquals(task.status, Task.Status.COMPLETE)
+            for assignment in task.assignments.all():
+                self.assertEquals(
+                    assignment.status, TaskAssignment.Status.SUBMITTED)
```
marcua: should we have a new status that's COMPLETED_AND_SKIPPED?

orchestra/utils/task_lifecycle.py
```diff
@@ -158,53 +159,6 @@ def _worker_certified_for_task(worker, task, role,
     return certified_for_task
 
 
-def _role_required_to_assign(task):
```
marcua: why were we able to eliminate this?

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
     return task
 
 
+@transaction.atomic
+def reassign_assignment(worker_id, assignment_id):
```
marcua, line 209: oh wow you really simplified things!:)

nice!  that old function was doing too many things

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+
+
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def reassign_assignment_api(request):
+    worker_username = json.loads(request.body.decode())['worker_username']
+    worker = Worker.objects.get(user__username=worker_username)
+    assignment_id = json.loads(request.body.decode())['assignment_id']
+    reassign_assignment(worker.id, assignment_id)
```
marcua: high-level: all of these apis should capture exceptions and include them in a returned json response, to parallel the other methods that do this

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+    return task
+
+
+@transaction.atomic
+def revert_task_to_datetime(task_id, revert_datetime, fake=False):
+    task = Task.objects.get(id=task_id)
+    audit = {'task': task, 'change': None, 'assignments': []}
+
+    assignments_after = (assignment_history(task)
+                         .filter(start_datetime__gt=revert_datetime)
```
marcua: two comments: 1) gt or gte, and 2) can we add to the revert tests assurances of what happens when a timedates are on a boundary condition?

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+                         .all())
+
+    if assignments_after.exists():
+        for assignment in assignments_after:
+            audit['assignments'].append({'assignment': assignment,
+                                         'change': 'deleted',
+                                         'snapshots': []})
+        if not fake:
+            # Delete assignments created after the revert datetime.
+            assignments_after.delete()
```
marcua: any time i see a delete, i get scared because of cascading deletes.  can you assure this doesn't cascade to other objects/models?  happy to explain tricks for doing this, but one simple one is to delete an assignment object in the admin and see whether it warns you about cascading deletes

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+        audit['change'] = 'deleted'
+    elif audit['assignments']:
+        if not fake:
+            _revert_task_status(task)
+        audit['change'] = 'reverted'
+
+    return audit
+
+
+@transaction.atomic
```
marcua: do you actually want this to be recursively atomic inside of the other revert?  if not, remove.  you typically do this when you want to be able to partially roll back (in the database transaction sense) a transaction

orchestra/tests/test_task_lifecycle.py
```diff
@@ -172,117 +177,246 @@ def test_worker_has_reviewer_status(self):
...
+                    'work_time_seconds': 0,
+                    'data': {
+                        snapshot['worker'].user.username: snapshot['type']
+                    },
+                })
+                assignment.save()
+
+            return task
+
+        def _test_reverted_task(datetime, status, num_assignments,
```
marcua: reminder: test the audit history

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+    for assignment in assignment_history(task).all():
+        assignment_audit = _revert_assignment_to_datetime(
+            assignment, revert_datetime, fake)
+        if assignment_audit:
+            audit['assignments'].append(assignment_audit)
+
+    # Delete task if it starts after revert_datetime
+    if task.start_datetime > revert_datetime:
+        if not fake:
+            task.delete()
```
marcua, line 348: same question about cascades

orchestra/utils/task_lifecycle.py
marcua, line 274: include docstring with short example audit format?  is it a list of audit records, or a nested structure?

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+
+
+@transaction.atomic
+def _revert_assignment_to_datetime(assignment, revert_datetime, fake=False):
+    audit = None
+    # Revert assignment to datetime by removing snapshots
+    gt_idx = None
+    snapshots = assignment.snapshots['snapshots']
+    for i, snapshot in enumerate(snapshots):
+        if parse_datetime(snapshot['datetime']) > revert_datetime:
```
marcua: sanity checking that this should be > and not >=

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+                for snapshot in snapshots[gt_idx:]
+            ]
+        }
+        if not fake:
+            assignment.snapshots['snapshots'] = snapshots[:gt_idx]
+            assignment.save()
+    return audit
+
+
+@transaction.atomic
```
marcua: same story w/ atomic

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+        assignment.status = TaskAssignment.Status.SUBMITTED
+        assignment.save()
+    create_subsequent_tasks(task.project)
+    return task
+
+
+@transaction.atomic
+def revert_task_to_datetime(task_id, revert_datetime, fake=False):
+    task = Task.objects.get(id=task_id)
+    audit = {'task': task, 'change': None, 'assignments': []}
```
marcua: change: None seems weird---`'unchanged'`?

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+        else:
+            latest_counter = last_active_assignment.assignment_counter
+            snapshot = last_active_assignment.snapshots['snapshots'][-1]
+            if snapshot['type'] == TaskAssignment.SnapshotType.REJECT:
+                # Task was last rejected back to previous worker
+                reverted_status = Task.Status.POST_REVIEW_PROCESSING
+                current_assignment_counter = latest_counter - 1
+                previous_assignment_counter = latest_counter
+
+            elif (snapshot['type'] == TaskAssignment.SnapshotType.SUBMIT
```
marcua: alternative: `snapshot['type'] in (TaskAssignment.SnapshotType.SUBMIT,TaskAssignment.SnapshotType.ACCEPT)` but go with whichever you feel is better

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+                    reverted_status = Task.Status.PENDING_REVIEW
+                else:
+                    reverted_status = Task.Status.REVIEWING
+                    previous_assignment_counter = latest_counter
+                    current_assignment_counter = latest_counter + 1
+
+    if reverted_status is None:
+        raise TaskStatusError('Task is malformed; cannot revert status.')
+
+    if current_assignment_counter:
```
marcua: if current_assignment_counter is 0, this will not fire---are you sure you don't want to explicitly test for None

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+        current_assignment = task.assignments.get(
+            assignment_counter=current_assignment_counter)
+        current_assignment.status = TaskAssignment.Status.PROCESSING
+        current_assignment.save()
+
+    if previous_assignment_counter:
```
marcua: same as above

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+                    current_assignment_counter = latest_counter + 1
+
+    if reverted_status is None:
+        raise TaskStatusError('Task is malformed; cannot revert status.')
+
+    if current_assignment_counter:
+        current_assignment = task.assignments.get(
+            assignment_counter=current_assignment_counter)
+        current_assignment.status = TaskAssignment.Status.PROCESSING
+        current_assignment.save()
```
marcua, line 479: if not fake

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+
+    if previous_assignment_counter:
+        previous_assignment = task.assignments.get(
+            assignment_counter=current_assignment_counter)
+        previous_assignment.status = TaskAssignment.Status.SUBMITTED
+        previous_assignment.save()
```
marcua, line 485: if not fake

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +228,182 @@ def assign_task(worker_id, task_id):
...
+
+    if not fake:
+        task.status = reverted_status
+        task.save()
+    return task
+
+
+def _get_last_active_assignment(task):
```
marcua: curiously, should this live in `task_properties.py`?

marcua: `#TODO(jrbotros): Consider actually stepping through the snapshots in order rather than relying on time, which might be unreliable.`

see: https://aphyr.com/posts/299-the-trouble-with-timestamps

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+def create_subsequent_tasks_api(request):
+    project_id = json.loads(request.body.decode())['project_id']
+    create_subsequent_tasks(Project.objects.get(id=project_id))
+
+
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def edit_slack_membership_api(request):
+    body = json.loads(request.body.decode())
```
marcua: remove all business logic for this function and place in another module

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+def complete_and_skip_task_api(request):
+    task_id = json.loads(request.body.decode())['task_id']
+    complete_and_skip_task(task_id)
+
+
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def assign_task_api(request):
+    worker_username = json.loads(request.body.decode())['worker_username']
```
marcua: remove all business logic for this function and place in another module

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+    assignment_id = json.loads(request.body.decode())['assignment_id']
+    reassign_assignment(worker.id, assignment_id)
+
+
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def revert_task_api(request):
+    body = json.loads(request.body.decode())
+    # Python datetime expects seconds, while javascript provides milliseconds
```
marcua: we should define the API Pythonically, and do any correction in javascript.

this is a minor nit, but the idea is that the api can be used by multiple end-users (unlikely) and so the server should set the standard that clients follow.

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+                del iteration['work_time']
+                assignment['iterations'].append(iteration)
+    return project_information
+
+
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def reassign_assignment_api(request):
+    worker_username = json.loads(request.body.decode())['worker_username']
```
marcua: i know why you did this (the latency api), but don't love it.  can we get the ID instead?  perhaps by passing the ID via the latency api?

it's ok, but not lovely, to punt this :)

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
...
+    worker = Worker.objects.get(user__username=worker_username)
```
marcua: ideally we wouldn't hit our data model in views, so even though it's a one-liner, relegate this query to a function in another module.

super-nitpicky of me...

orchestra/views.py
```diff
@@ -183,6 +204,128 @@ def submit_task_assignment(request):
         raise BadRequest(e)
 
 
+@json_view
+@login_required
+@user_passes_test(is_project_admin)
+def project_management_information_api(request):
+    project_id = json.loads(request.body.decode())['project_id']
```
marcua: remove all business logic for this function and place in another module

orchestra/views.py
```diff
@@ -1,39 +1,60 @@
...
+from orchestra.utils.task_lifecycle import revert_task_to_datetime
+from orchestra.utils.task_lifecycle import assign_task
 from orchestra.utils.task_lifecycle import worker_assigned_to_rejected_task
 from orchestra.utils.task_lifecycle import worker_assigned_to_max_tasks
 from orchestra.utils.task_lifecycle import worker_has_reviewer_status
+from orchestra.utils.task_lifecycle import end_project
 
 import logging
 
 logger = logging.getLogger(__name__)
```
marcua, line 34: high-level:
- create business logic functions that live outside of this module
- ideally test all endpoints you've introduced rather than the business logic implementations they call, so we test as close to end-to-end as possible
- a lot of the new methods are missing exception handling & status JSON response creation

marcua:
Back over to you!  I left `high-level` comments in various places where i try to capture bigger-picture ideas.

Thanks to @thisisdhaas for talking through the javascript module stuff!

orchestra/analytics/latency.py
jrbotros: discussed offline with @marcua—will add an issue summarizing our thoughts

orchestra/analytics/latency.py
jrbotros: Thoughts summed up in #78

jrbotros: :white_check_mark:

orchestra/analytics/latency.py
jrbotros: Thoughts summed up in #78

jrbotros: :white_check_mark:

orchestra/static/orchestra/common/js/orchestra_api.js
jrbotros: Not right now—was planning on adding in the rest in a separate PR, let me know if it should go in this one.

orchestra/static/orchestra/common/js/orchestra_api.js
jrbotros, line 2: :white_check_mark:

orchestra/api_urls.py
jrbotros: :white_check_mark:

orchestra/static/orchestra/common/js/orchestra_api.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/css/project.scss
jrbotros: Maybe because they're SVG-specific and that makes the CSS linter sad? I copied `shape-rendering` from a Bostock example and path is definitely supported so I don't think there's an issue.

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: Leftovers from debugging, oops :flushed:

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: I try :star2:

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: I've tried to get rid of them all, let me know if you see any stray ones!

jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/data_service.js
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: Yeah, I definitely think so—do you think that can go in a separate PR?

orchestra/utils/task_lifecycle.py
jrbotros, line 209: :wink:

jrbotros, line 209: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: Oops.

jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros, line 479: I'm going to take out the `fake` argument to this function, since we're not actually calling it unless `fake` is false in the parent revert method.

orchestra/utils/task_lifecycle.py
jrbotros, line 485: See above.

jrbotros, line 485: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros, line 479: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: Nope! http://stackoverflow.com/a/13497741

thisisdhaas:
Friendly idea: display the workflow slug / workflow version slug in the interface somewhere?

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: I separated things out—take a look and let me know!

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: Dead and gone.

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/data_service.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/data_service.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: I guess there's not a situation where we would roll back the assignment and not the whole task? That was the original intention, but it doesn't really make sense in retrospect.

jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: The interface requires the user to provide the name of the worker to reassign, so we have to take in the username here.

jrbotros: :white_check_mark:

orchestra/static/orchestra/common/js/orchestra_api.js
jrbotros: Unnecessary, taken out!

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: Not anymore!

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: Planning to hold this off for a later PR.

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: See above!

jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/data_service.js
jrbotros: Fixed!

And it's not actually task data, just metadata about the visualization (for now, just whether the task is expanded or not).

jrbotros: :white_check_mark:

jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

jrbotros: Talked about offline; single object retrievals can live within the API methods as long as nothing else is modified.

jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: See above.

jrbotros: :white

jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: Right now, the function is specified in the interface as reverting to before the specified datetime, so > makes sense. Maybe we should rename this `revert_task_to_before_datetime`?

orchestra/utils/task_lifecycle.py
jrbotros: Right now, the function is specified in the interface as reverting to before the specified datetime, so gt makes sense. Maybe we should rename this `revert_task_to_before_datetime`?

(This particular piece of logic was moved, but the point should stand.)

jrbotros: 2) Boundary tests still need to be added

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/common/css/common.scss
jrbotros, line 30: :white_check_mark: (mostly, a few minor bugs will be left outstanding)

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros: :white_check_mark:

orchestra/views.py
jrbotros, line 34: :white_check_mark:

orchestra/static/orchestra/common/js/orchestra_api.js
marcua: added https://github.com/unlimitedlabs/orchestra/issues/79

marcua:
reminder to uncomment all `strict` lines

---

@jrbotros
:white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
marcua: added https://github.com/unlimitedlabs/orchestra/issues/80

orchestra/tests/test_task_lifecycle.py
marcua: added https://github.com/unlimitedlabs/orchestra/issues/81

orchestra/project_api/serializers.py
```diff
@@ -67,7 +66,18 @@ def get_status(self, obj):
         return dict(Task.STATUS_CHOICES).get(obj.status, None)
 
     def get_latest_data(self, obj):
```
marcua, line 72: minor nitpick:

```
    """
    Return latest input data for a specified task.

    Args:
        task (orchestra.models.Task):
            The task object for which to retrieve data.

    Returns:
        latest_data (str):
            A serialized JSON blob containing the latest input data.
    """
```

jrbotros, line 72: :white_check_mark:

orchestra/static/orchestra/common/js/orchestra_api.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/css/project.scss
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/project/js/controllers.js
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros, line 171: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros, line 172: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros, line 274: When we rethink how we handle iterations throughout the code I'll take another pass at the revert stuff!

jrbotros, line 274: :white_check_mark: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: I don't think there's a ton of benefit, since the initial status of the task doesn't factor in anywhere.

jrbotros: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros, line 20: :white_check_mark:

orchestra/tests/test_task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros: :white_check_mark:

orchestra/utils/task_lifecycle.py
jrbotros, line 348: :white_check_mark:

orchestra/utils/task_lifecycle.py
```diff
@@ -275,6 +205,298 @@ def assign_task(worker_id, task_id):
...
+    # TODO(jrbotros): The revert methos should "peel off" snapshots,
+    # rather than deleting them in bulk and recalculating the assignment
+    # and task statuses; this logic needs to be cleaned up.
+    for assignment in task.assignments.all():
+        if assignment.status == TaskAssignment.Status.SUBMITTED:
+            latest_snapshot = assignment.snapshots['snapshots'][-1]
+            assignment.in_progress_task_data = latest_snapshot['data']
+            assignment.save()
+
+    task.status = reverted_status
```
marcua: what if this is the task with the status that's SUBMITTED?

orchestra/tests/helpers/fixtures.py
```diff
@@ -453,3 +461,178 @@ def fix_datetimes(snapshots, new_datetimes):
     third_assignment.save()
 
     return task
+
+
+def setup_complete_task(test_case, times):
+    # Microseconds are truncated when manually saving models
+    # test_start = test_start.replace(microsecond=0)
```
marcua: halp

orchestra/tests/helpers/fixtures.py
```diff
@@ -453,3 +461,178 @@ def fix_datetimes(snapshots, new_datetimes):
...
+    test_case.assertEquals(task.status, Task.Status.COMPLETE)
+    test_case.assertEquals(task.assignments.count(), 2)
+    for assignment in task.assignments.all():
+        test_case.assertEquals(
+            assignment.status, TaskAssignment.Status.SUBMITTED)
+        test_case.assertEquals(len(assignment.snapshots['snapshots']), 2)
+    return task
+
+
+def test_audit(test_case, audit, num_assignments,
```
marcua: not a fixture, should be a _private_method on the testcase

orchestra/tests/helpers/fixtures.py
```diff
@@ -453,3 +461,178 @@ def fix_datetimes(snapshots, new_datetimes):
...
+        else:
+            assignment['change'] == 'reverted'
+        for j, snapshot in enumerate(assignment['snapshots']):
+            if j > num_snapshots - 1:
+                test_case.assertEquals(snapshot['change'], 'deleted')
+            else:
+                test_case.assertEquals(snapshot['change'], 'unchanged')
+
+
+def test_reverted_task(test_case, times, datetime, status, num_assignments,
```
marcua: not a fixture, should be a _private_method on the testcase

orchestra/tests/helpers/fixtures.py
```diff
@@ -453,3 +461,178 @@ def fix_datetimes(snapshots, new_datetimes):
...
+        reverse('orchestra:orchestra:project_management:revert_task'),
+        json.dumps({
+            'task_id': task.id,
+            # Convert datetime to timestamp
+            'revert_datetime': time.mktime(datetime.timetuple()),
+            'fake': True
+        }),
+        content_type='application/json')
+    test_case.assertEquals(response.status_code, 200)
+    # fake_audit = json.loads(response.content.decode('utf-8'))
```
marcua: hey

jrbotros: :whit

jrbotros: :white_check_mark:

orchestra/tests/helpers/fixtures.py
jrbotros: :white_check_mark:

orchestra/tests/helpers/fixtures.py
jrbotros: :white_check_mark:

orchestra/tests/helpers/fixtures.py
jrbotros: :white_check_mark:

PR #75, opened 2015-11-29
Adds full Project API support for workflows in DB.

thisisdhaas:


docs/source/example_use.rst
```diff
@@ -49,7 +49,7 @@ If you haven't followed the :doc:`getting started guide <getting_started>` to
 set up Orchestra yet, you should do that now. Also, make sure that
 ``'journalism_workflow'`` is in your ``INSTALLED_APPS`` setting, and that you
 have loaded the workflow into the database (``python manage.py loadworkflow
-journalism_workflow journalism_v1``).
+journalism_workflow v1``).
```
thisisdhaas, line 52: I left this as `<APP_LABEL> <WORKFLOW_VERSION_SLUG>` intentionally to differentiate it from the `<WORKFLOW_SLUG>/<WORKFLOW_VERSION_SLUG>` format accepted by the `loadworkflowsampledata` command. I'm open to suggestions for a cleaner way of highlighting this difference.

orchestra/project_api/views.py
```diff
@@ -74,6 +73,11 @@ def project_details_url(request):
...
-    workflows = WorkflowVersion.objects.all()
-    workflow_choices = {w.slug: w.name for w in workflows}
-    return {'workflows': workflow_choices}
+    workflows = {
+        w.slug: {
+            'name': w.name,
+            'versions': {v.slug: v.name for v in w.versions.all()}
+        }
+        for w in Workflow.objects.all()
+    }
```
thisisdhaas, line 88: I think this should probably just spit out the `workflow.json` and `version.json` manifest for all installed workflows/versions, but this is a simpler change for now. Thoughts? (I'd also consider adding more fields into this gradually, `description` and `certifications` are obvious candidates)

journalism_workflow/v1/version.json
```diff
@@ -1,5 +1,5 @@
 {
-  "slug": "journalism_v1",
+  "slug": "v1",
```
thisisdhaas, line 2: Calling out that I'm renaming the journalism/simple_workflow version slugs here. This will cause confusion for anyone who has already loaded these workflows into the DB, but I don't think a) any such person exists other than me and b) it's a big issue, since these are demo workflows.

docs/source/example_use.rst
marcua, line 52: i like calling out the difference.  one other way to do this is with `--app` and `--workflow-version` flags. the only reason to have `workflow/version` format is when we want to allow lists of those rather than individual values.  what do you think?

orchestra/project_api/views.py
marcua, line 88: maybe let's do `{v.slug: {'name': v.name, 'description': v.description}}` for now.  this response is two-part:
- make the inner value a dict, for consistency with the parent element, and
- description seems useful for display.  i feel less strongly about this second point than i do abotu the first point :)

journalism_workflow/v1/version.json
marcua, line 2: sanity check that the docs are appropriately updated?

docs/source/api.rst
```diff
@@ -106,14 +108,26 @@ All requests must be signed using `HTTP signatures <http://tools.ietf.org/html/d
...
 
       {
-        "workflows": {
-                       "reporting": "A sample workflow for the newsroom."
-                     }
+          "workflows": {
+	      "journalism": {
+	          "name": "Journalism Workflow",
+		  "versions": {
+		      "v1": "Journalism Workflow Version 1",
```
marcua: note to update this when you update the inner element to be a dict

docs/source/api.rst
```diff
@@ -106,14 +108,26 @@ All requests must be signed using `HTTP signatures <http://tools.ietf.org/html/d
...
+		      "v2": "Journalism Workflow Version 2"
+                  }
+	      },
+	      "simple_workflow": {
+	          "name": "Simple Workflow",
+		  "versions": {
+		      "v1": "Simple Workflow Version 1"
```
marcua: same note

docs/source/example_use.rst
```diff
@@ -60,7 +60,7 @@ documentation <https://developers.google.com/drive/web/delegation>`_.
 
 Next, make sure you have the journalism workflow sample data installed by
 running (if you haven't already) ``python manage.py loadworkflowsampledata
-journalism journalism_v1``. This will create the following accounts:
+journalism/v1``. This will create the following accounts:
```
marcua, line 63: see my note about `--flags` when we're not passing in a list of these

docs/source/example_use.rst
```diff
@@ -414,4 +414,4 @@ signature, and might look like (for example)::
 
 Once that function is defined, sample data can be loaded with::
 
-  python manage.py loadworkflowsampledata <WORKFLOW_SLUG> <VERSION_SLUG>
+  python manage.py loadworkflowsampledata <WORKFLOW_SLUG>/<WORKFLOW_VERSION>
```
marcua, line 417: same note

docs/source/getting_started.rst
```diff
@@ -132,17 +132,17 @@ To load the workflows, run::
 Each of our example workflows provides a set of sample users already configured
 with proper certifications. To load them, run::
 
-    python manage.py loadworkflowsampledata <WORKFLOW_SLUG> <WORKFLOW_VERSION>
+    python manage.py loadworkflowsampledata <WORKFLOW_SLUG>/<WORKFLOW_VERSION>
```
marcua, line 135: same note

orchestra/management/commands/loadworkflowsampledata.py
```diff
@@ -19,29 +19,35 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            'workflow_slug',
-            help='The unique identifier of the workflow.')
-        parser.add_argument(
-            'version_slug',
-            help='The version of the workflow to load.')
+            'workflow_version',
+            help=('The unique identifier of the workflow version (formatted '
```
marcua, line 23: same note :)

journalism_workflow/v1/version.json
thisisdhaas, line 2: Super updated.

orchestra/project_api/views.py
thisisdhaas, line 88: Yeah, the object feels pretty empty with just 'name' in it. Donesies.

docs/source/api.rst
thisisdhaas: up-drated.

docs/source/api.rst
thisisdhaas: updated!

docs/source/example_use.rst
thisisdhaas, line 52: Discussed, leaving as is.

docs/source/example_use.rst
thisisdhaas, line 63: again, discussed and leaving as is.

docs/source/example_use.rst
thisisdhaas, line 417: same answer.

docs/source/getting_started.rst
thisisdhaas, line 135: :ballot_box_with_check:

orchestra/management/commands/loadworkflowsampledata.py
thisisdhaas, line 23: :checkered_flag: :ballot_box_with_check:

marcua:
lgtm!  :shipit:

PR #77, opened 2015-11-30
Refactor task services

jrbotros:
Moves `autoSave` functionality into its own Angular service!

### DIFF
--- a/orchestra/static/orchestra/task/js/autosave.js
+++ b/orchestra/static/orchestra/task/js/autosave.js
@@ -1,0 +1,90 @@
+(function() {
+  'use strict';
+
+  var serviceModule =  angular.module('orchestra.task.services');
+
+  serviceModule.factory('autoSaveTask', function($rootScope, $timeout, $http, orchestraService) {
+    var autoSaveTask = new function() {
+      var autoSaver = this;
+
+      autoSaver.setup = function($scope, taskId, taskData) {
+        autoSaver.saveError = false;
+        autoSaver.saving = false;
+        autoSaver.timeout = 10000;
+        autoSaver.taskId = taskId;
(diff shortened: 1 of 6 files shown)


orchestra/static/orchestra/task/js/autosave.js
```diff
@@ -0,0 +1,87 @@
...
+            if (!confirm('Your latest changes haven\'t been saved.\n\n' +
+                  'Are you sure you want to leave this page?')) {
+              e.preventDefault();
+            }
+          }
+        });
+      }
+
+      autoSaver.schedule = function() {
+        if (!autoSaver.autoSaveTimer && !autoSaver.is_read_only) {
```
marcua: likely from the scope and not from autosaver?

let's manually test
- autosave works
- save works
- no autosave on readonly tasks

orchestra/static/orchestra/task/js/autosave.js
```diff
@@ -0,0 +1,87 @@
+(function() {
```
marcua, line 1: nitpick: this is a taskSaver with autosaving functionality.  so perhaps calling it a taskSaver and task_save.js might make more sense.

but that's just one person's opinion :)

orchestra/static/orchestra/task/js/autosave.js
jrbotros: :white_check_mark: on both counts!

orchestra/static/orchestra/task/js/autosave.js
jrbotros, line 1: I feel like it's an autoSaver which wraps saving functionality—going to leave it as is, I think :blush:

PR #85, opened 2015-12-07
Optionally require fields in task view

jrbotros:
Provides a directive to wrap required fields in task views.
- The wrapped HTML should contain an input element bound with ngModel.
- Optionally provide a class to add to the directive wrapper on error with `data-error-class="error-class"` on the directive element; otherwise, a default is provided for some input types.

screen shot 2015-12-07 at 6 39 46 pm

### DIFF
--- a/orchestra/static/orchestra/task/js/controllers.js
+++ b/orchestra/static/orchestra/task/js/controllers.js
@@ -6,7 +6,8 @@
     .controller('TaskController', TaskController);
 
   function TaskController($location, $scope, $routeParams, $http, $rootScope,
-                          $modal, $timeout, autoSaveTask, orchestraService) {
+                          $modal, $timeout, autoSaveTask, orchestraService,
+                          requiredFields) {
     var vm = this;
     vm.taskId = $routeParams.taskId;
     vm.taskAssignment = {};
@@ -30,6 +31,7 @@
           vm.projectFolderExternalUrl = orchestraService.googleUtils.folders.externalUrl(projectFolderId)
 
           if (!vm.is_read_only) {
+            requiredFields.setup(vm);
             $scope.$watch('vm.taskAssignment.task.data', function(newVal, oldVal) {
               // Ensure save fired at initialization
               // [http://stackoverflow.com/a/18915585]
@@ -72,7 +74,12 @@
 
     vm.confirmSubmission = function(command, totalSeconds) {
       vm.submitting = true;
-      orchestraService.signals.fireSignal('submit.before');
+      if (orchestraService.signals.fireSignal('submit.before') === false) {
+        // If any of the registered signal handlers returns false, prevent
+        // submit.
+        vm.submitting = false;
+        return;
+      };
       $http.post('/orchestra/api/interface/submit_task_assignment/',
                  {'task_id': vm.taskId, 'task_data': vm.taskAssignment.task.data,
                   'command_type': command, 'work_time_seconds': totalSeconds})
--- a/orchestra/static/orchestra/task/js/required_fields.js
+++ b/orchestra/static/orchestra/task/js/required_fields.js
@@ -1,0 +1,104 @@
+(function() {
+  'use strict';
+
+  var serviceModule =  angular.module('orchestra.task.services');
+
+  /**
+   * Service to keep track of all required fields in a task view and
+   * validate them on submit.
+   */
+  serviceModule.factory('requiredFields', function($rootScope, orchestraService) {
+    var requiredFields = {
+      fields: [],
+      invalid: [],
+      setup: function(data) {
+        /**
+         * Sets up the base data on which to validate fields.
+         */
+        this.data = data;
+      },
+      require: function(fields) {
+        /**
+         * Sets a field as required. Fields are specified by
+         * dot-delimited key strings (e.g., `key0.key1.key2`).
+         */
+        this.fields = this.fields.concat(fields)
+      },
+      validate: function() {
+        /**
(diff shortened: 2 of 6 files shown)


orchestra/static/orchestra/task/js/controllers.js
```diff
@@ -72,7 +72,12 @@
 
     vm.confirmSubmission = function(command, totalSeconds) {
       vm.submitting = true;
-      orchestraService.signals.fireSignal('submit.before');
+      if (orchestraService.signals.fireSignal('submit.before') === false) {
```
marcua: style points---they just clicked submit and are at the bottom of the task, so they might think something broke because hte error is at the top.  throw a message or tooltip on the button saying something went wrong?

totally OK (and encouraged) for you to turn this into a `TODO(jrbotros)`

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Service to keep track of all required fields in a task view and
+   * validate them on submit.
+   */
+  serviceModule.factory('requiredFields', function($rootScope, orchestraService) {
+    var requiredFields = {
+      fields: [],
+      invalid: [],
+      setup: function(data) {
```
marcua: this function is not called

marcua, line 14: who calls this function, and what is data?

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+        this.data = data;
+      },
+      require: function(fields) {
+        this.fields = this.fields.concat(fields)
+      },
+      validate: function() {
+        var requiredFields = this;
+        requiredFields.invalid = [];
+        this.fields.forEach(function(field) {
+          var obj = requiredFields.data;
```
marcua, line 38: what is obj?  better name

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+          var keys = field.split('.');
+          for (var i in keys) {
+            var key = keys[i];
+            obj = obj[key];
```
marcua, line 42: why override obj?

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+      setup: function(data) {
+        this.data = data;
+      },
+      require: function(fields) {
+        this.fields = this.fields.concat(fields)
+      },
+      validate: function() {
+        var requiredFields = this;
+        requiredFields.invalid = [];
+        this.fields.forEach(function(field) {
```
marcua: this->requiredFields

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+          for (var i in keys) {
+            var key = keys[i];
+            obj = obj[key];
+            if (!obj) {
+              requiredFields.invalid.push(field)
+              break;
+            }
+          }
+        })
+        $rootScope.$broadcast('orchestra:task:validatedFields');
```
marcua, line 49: curiously, if there's a broadcast mechanism, did we need to have our own pre/post signals?  they're useful now because they can return false, but i wonder if we needed them otherwise?

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -0,0 +1,90 @@
...
+      },
+      require: function(fields) {
+        this.fields = this.fields.concat(fields)
+      },
+      validate: function() {
+        var requiredFields = this;
+        requiredFields.invalid = [];
+        this.fields.forEach(function(field) {
+          var obj = requiredFields.data;
+          var keys = field.split('.');
```
marcua, line 39: why do you have to split?  i probably don't understand what a field looks like, which results in some of the confusion in these comments :)

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 14: This function was originally being called by custom task view controllers, but I think it makes sense to set this up in the base task controller instead.

Data is the base object which field indicators correspond to. I'll add a comment.

orchestra/static/orchestra/task/js/required_fields.js
jrbotros: :white_check_mark:

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 49: We wouldn't have before, oops. Now useful though :)

jrbotros, line 49: :white_check_mark:

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 38: See if comments clear things up!

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 42: See if comments clear things up!

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 39: See if comments clear things up!

marcua:
lgtm!

PR #89, opened 2015-12-11
Surface directives from multiple workflow versions

jrbotros:


thisisdhaas:
oh whoops--this looks like something I should have taken care of in 96208ce: my bad!

thisisdhaas:
LGTM

PR #90, opened 2015-12-12
Google analytics support

marcua:


example_project/example_project/orchestra_settings.py
```diff
@@ -57,6 +57,11 @@ def setup_orchestra(settings_module_name):
         ('text/x-scss', 'django_libsass.SassCompiler'),
     )
 
+    # Pass global variables (e.g., Google Analytics key) to templates with
+    # this context processor.
+    settings.TEMPLATES[0]['OPTIONS']['context_processors'].append(
+        'orchestra.context_processors.global_settings')
```
thisisdhaas: Two comments:
- Is this global_settings trick something that others have done / a common design pattern in Django? It feels a bit odd to me...
- https://docs.djangoproject.com/en/1.9/ref/settings/#templates suggests that the default setting is an empty list--we probably need to check for that and add a default entry ourselves.

orchestra/context_processors.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+
+
+def global_settings(request):
```
thisisdhaas: yeah, I'm just worried that this will get abused in the future... Is there any way to associate a context_processor with a single template or a list of templates? Or is there relatively minor syntax to let us call something like `render('path/to/template', extra_context_processors=[global_settings,])`?

example_project/example_project/orchestra_settings.py
marcua: Added more sanity-checking as per an offline conversation

orchestra/context_processors.py
marcua: made Google Analytics-specific as per an offline conversation

orchestra/context_processors.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+
+
+def google_analytics(request):
+    """
+    Make any variables you'd like global access to in templates available here.
+    """
```
thisisdhaas, line 7: Not the right docstring anymore :-)

example_project/example_project/orchestra_settings.py
```diff
@@ -186,6 +186,24 @@ def setup_orchestra(settings_module_name):
...
+            'orchestra.context_processors.google_analytics')
+    except:
+        raise ValueError(
+            "Expected settings.TEMPLATES to contain a single DjangoTemplates "
+            "entry with `['OPTIONS']['context_processors']` in which to "
+            "place a Google Analytics context processor.  If your template "
+            "setup is more complex, please manually add "
+            "'orchestra.context_processors.google_analytics' to "
+            "the appropriate context processor list.")
+
```
thisisdhaas, line 210: maybe worth sticking in an `install_google_analytics()` function like `install_admin()` below, to keep things clean?

Your call.

thisisdhaas:
LGTM if you remove the out-of-date docstring, consider my other comment, and sanity-check that the docs don't need to change because we're playing with settings.

---

AM: made a quick update to docs

example_project/example_project/orchestra_settings.py
marcua, line 210: done!

orchestra/context_processors.py
marcua, line 7: done!

PR #91, opened 2015-12-14
Retool required field functionality

jrbotros:
Genericizes field validation to use the HTML elements themselves (rather than their ng-model attribute) to support fields nested inside directives. Also allows defining custom field types and registering corresponding validator functions.

### DIFF
--- a/orchestra/static/orchestra/task/js/required_fields.js
+++ b/orchestra/static/orchestra/task/js/required_fields.js
@@ -9,45 +9,77 @@
    */
   serviceModule.factory('requiredFields', function($rootScope, orchestraService) {
     var requiredFields = {
-      fields: [],
+      fields: {},
       invalid: [],
+      validators: {
+        'input-checkbox': [
+          function(elem) {
+            return elem.checked;
+          }
+        ],
+        'input-text': [
+          function(elem) {
+            return elem.value && elem.value.length > 0;
+          }
+        ],
+      },
       setup: function(data) {
         /**
          * Sets up the base data on which to validate fields.
          */
         this.data = data;
       },
-      require: function(fields) {
+      require: function(fieldType, field) {
         /**
-         * Sets a field as required. Fields are specified by
-         * dot-delimited key strings (e.g., `key0.key1.key2`).
+         * Sets a field as required. Fields are HTML elements to be
+         * checked by one or more validators according to their field
+         * type.
          */
-        this.fields = this.fields.concat(fields)
+        if (this.fields[fieldType] === undefined) {
+          this.fields[fieldType] = [field];
+        }
+        else {
+          this.fields[fieldType].push(field);
+        }
       },
       validate: function() {
         /**
-         * Validates required fields. For the required field
-         * `key0.key1.key2` to be valid, `this.data[key0][key1][key2]`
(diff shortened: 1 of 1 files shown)


orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -9,45 +9,73 @@
...
+      require: function(fieldType, field) {
         /**
-         * Sets a field as required. Fields are specified by
-         * dot-delimited key strings (e.g., `key0.key1.key2`).
+         * Sets a field as required. Fields are HTML elements to be
+         * checked by one or more validators according to their field
+         * type.
          */
-        this.fields = this.fields.concat(fields)
+        if (Object.keys(this.fields).indexOf(fieldType) >= 0) {
```
thisisdhaas: is this the standard way to check for a key on an associative array? Doesn't just `if (this.fields[fieldType] != undefined)` work?

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -9,45 +9,73 @@
...
-              requiredFields.invalid.push(field)
-              break;
+        for (var fieldType in requiredFields.fields) {
+          var validators = requiredFields.validators[fieldType];
+          var fields = requiredFields.fields[fieldType];
+          fields.forEach(function(field) {
+            var success = true;
+            validators.forEach(function(validator) {
+              success = success && validator(field);
+            })
```
thisisdhaas, line 63: should we catch the case where someone requires validation for a non-existent field type? e.g., fields contains `'input-radio'` but we don't support it? Here, validators will be undefined and `forEach` will barf confusingly, I think.

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -9,45 +9,73 @@
...
+        }
         $rootScope.$broadcast('orchestra:task:validatedFields');
         return requiredFields.invalid.length === 0;
+      },
+      registerValidator: function(fieldType, validator) {
+        /**
+         * Register a validator function to the given field type.
+         */
+        var requiredFields = this;
+        if (Object.keys(requiredFields.validators).indexOf(fieldType) >= 0) {
```
thisisdhaas: same comment.

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -9,45 +9,73 @@
...
+          validator[fieldType].push(validator);
```
thisisdhaas: `validator`? Or `requiredFields.validators`?

orchestra/static/orchestra/task/js/required_fields.js
```diff
@@ -9,45 +9,73 @@
...
+        }
+        else {
+          validator[fieldType] = [validator];
```
thisisdhaas: Same comment.

orchestra/static/orchestra/task/js/required_fields.js
jrbotros, line 63: :white_check_mark:

thisisdhaas:
LGTM after a squash/rebase.

PR #93, opened 2015-12-16
Remove project drive folder from base task interface

jrbotros:
This should be surfaced by step interfaces so it can be put in context.


marcua:
lgtm

PR #95, opened 2015-12-17
Automatically create team messages, make their iframe embeddable, and update docs

marcua:


docs/source/api.rst
```diff
@@ -55,7 +54,7 @@ All requests must be signed using `HTTP signatures <http://tools.ietf.org/html/d
               "id": 123,
               "short_description": "Project Description",
               "priority": 10,
-              "review_document_url": "http://review.document.url",
+              "team_messages_url": "http://review.document.url",
```
thisisdhaas, line 57: shouldn't that be documented above then?

EDIT: ah I see--we're not passing this in anymore, correct?

orchestra/google_apps/convenience.py
```diff
@@ -21,6 +21,7 @@
 logger = logging.getLogger(__name__)
 _image_mimetype_regex = re.compile('(image/(?:jpg|jpeg|gif|png))',
                                    re.IGNORECASE)
+TEAM_MESSAGES_TEMPLATE_ID = '1d0kIgq8G_Su6j5abP-tP6yJ2sp-sFDk6vZiREil0_70'
```
thisisdhaas, line 24: Looks like the file is public-writable--make read-only?

thisisdhaas: LGTM

thisisdhaas:
LGTM

docs/source/api.rst
marcua, line 57: correct on the second question, since it gets automatically created

orchestra/google_apps/convenience.py
marcua, line 24: Done, thanks!

PR #98, opened 2015-12-18
Fix bug in project management view

jrbotros:


orchestra/analytics/latency.py
```diff
@@ -55,12 +55,17 @@ def work_time_df(projects, human_only=True, complete_tasks_only=True):
             get_project_information(project.id),
             human_only, complete_tasks_only)
         for project in projects))
-    df = DataFrame(time_data)
+
+    # Ensure that even an empty DataFrame has the correct columns
+    columns = ['assignment_level', 'calendar_time', 'end_datetime',
+               'iteration', 'project_description', 'project_id',
+               'start_datetime', 'task_id']
```
thisisdhaas: are these names spelled out elsewhere ever? if so, should they be a module level var `DATAFRAME_COLUMNS` somewhere?

orchestra/interface_api/project_management/project_management.py
```diff
@@ -20,7 +20,6 @@ def project_management_information(project_id):
     project = Project.objects.get(id=project_id)
     df = work_time_df([project],
                       human_only=False, complete_tasks_only=False)
-    df['end_datetime'] = df['start_datetime'] + df['calendar_time']
     project_information = get_project_information(project.id)
```
thisisdhaas, line 23: why is this not necessary anymore?

thisisdhaas:
LGTM

orchestra/analytics/latency.py
marcua: if you do create a var, zip it with https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/analytics/latency.py#L81 so we don't repeat the column names a bunch of times

orchestra/interface_api/project_management/project_management.py
jrbotros, line 23: Was unnecessary to begin with :)

PR #100, opened 2015-12-22
Bump angular to v1.4.8

jrbotros:
To use the latest `ngOptions` directive and keep us up to date. Looked through the changelog and there aren't any significant breaking changes, so we should be good!


marcua:
lgtm!

PR #101, opened 2015-12-23
Make the "End Project" button more informative

marcua:
Users who see the old label are tempted to end a project in some clean way.


thisisdhaas:
LGTM

PR #102, opened 2016-01-19
Ensure certifications for entry_level are migrated before reviewer

jrbotros:
The command currently doesn't guarantee a migration order, so `reviewer` certifications are sometimes created before their `entry_level` counterparts (raising a `ModelSaveError`).


marcua:
lgtm

PR #104, opened 2016-02-24
Override django-registration RegisterView to emit custom orchestra signal

joshblum:


orchestra/signals.py
```diff
@@ -1,9 +1,12 @@
 from django.dispatch import receiver
-from registration.signals import user_registered
+from django.dispatch import Signal
 
 from orchestra.models import Worker
 
+# A user has activated his or her account.
+orchestra_user_registered = Signal(providing_args=["user", "request"])
```
marcua: single-quotes

orchestra/views.py
```diff
@@ -187,3 +193,49 @@ def submit_task_assignment(request):
 # A simple status endpoint for things like health checks, etc.
 def status(request):
     return HttpResponse('OK')
+
+
+class OrchestraRegistrationView(RegistrationView):
```
marcua, line 198: is this impl essentially the same as RegistrationView except we do `signals.orchestra_user_registered.send` at the end?  if so, can we make our implementation call the parent class' implementation?

joshblum, line 198: I wanted to, but the parent class emits the signal at the end of it. Would this be a problem if another project was using registration package and listening to the signal but didn't want to get notified by orchestra users?

marcua, line 198: ok cool---let's leave as you have it :)

marcua:
lgtm assuming you've manually tested that registration still works

joshblum:
screen shot 2016-02-24 at 11 22 48
 yallah :)

PR #135, opened 2016-04-04
Add iteration model in preparation for removing snapshots

jrbotros:


orchestra/migrations/0025_iteration.py
```diff
@@ -0,0 +1,29 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='Iteration',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('start_datetime', models.DateTimeField(default=django.utils.timezone.now)),
+                ('end_datetime', models.DateTimeField(blank=True, null=True)),
+                ('action', models.IntegerField(blank=True, choices=[(0, 'Submit'), (1, 'Accept'), (2, 'Reject')], null=True)),
+                ('data', jsonfield.fields.JSONField(blank=True, default={})),
+                ('assignment', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='iterations', to='orchestra.TaskAssignment')),
```
jrbotros, line 26: Should I be worried about this? Haven't seen `on_delete` autogenerated before.

jrbotros, line 26: Update: apparently it will be a required argument for `ForeignKey` in Django 2.0.

orchestra/models.py
```diff
@@ -461,6 +461,43 @@ def __str__(self):
             str(self.task), self.assignment_counter, str(self.worker))
 
 
+class Iteration(models.Model):
```
marcua, line 464: If TaskAssignment attaches to Task, should this be called TaskAssignmentIteration or TaskIteration or AssignmentIteration?

orchestra/models.py
```diff
@@ -461,6 +461,43 @@ def __str__(self):
...
+
+    Attributes:
+        start_datetime (datetime.datetime):
+            The time the task was assigned for this iteration.
+        end_datetime (datetime.datetime):
+            The time the iteration was completed. Will be None if the iteration
+            is currently in progress.
+        assignment (orchestra.models.TaskAssigment):
+            The task assignment to which the iteration belongs.
+        action (orchestra.models.Iteration.Action):
```
marcua: idea:  if we made it non-nullable, we could call it a `status` or `state`, which could be one of `PROCESSING, SUBMITTED, ACCEPTED, REJECTED`

if we eventually implement no deletions on data, we could also have an `INVALIDATED` state or somesuch

orchestra/models.py
```diff
@@ -461,6 +461,43 @@ def __str__(self):
...
+        SUBMIT = 0
+        ACCEPT = 1
+        REJECT = 2
+
+    ACTION_CHOICES = (
+        (Action.SUBMIT, 'Submit'),
+        (Action.ACCEPT, 'Accept'),
+        (Action.REJECT, 'Reject'))
+
+    start_datetime = models.DateTimeField(default=timezone.now)
```
marcua, line 495: chat w/ @joshblum about whether this should be `created_at`

orchestra/models.py
jrbotros, line 464: I really don't like this naming practice, so I was hoping to break it subtly :)

I would prefer that it be called iteration, but would prefer `AssignmentIteration` if you think we should continue to follow that convention.

orchestra/models.py
jrbotros: `REJECTED` sounds like the iteration itself was rejected, when in reality it just means that the worker is passing it back down the hierarchy. I do like the idea of having processing and invalidated statuses, though.

orchestra/models.py
jrbotros, line 495: Right now projects, tasks, assignments, workers all have fields called `start_datetime`—probably easier to be consistent for now and change them all in one go if we decide to?

(I know I advocated for inconsistency above, don't hate me too much for my hypocrisy :yum:)

orchestra/models.py
marcua, line 464: OK let's leave as-is

orchestra/models.py
marcua, line 495: :+1:

orchestra/models.py
jrbotros, line 464: :white_check_mark:

orchestra/models.py
jrbotros: Decided offline on:
`REJECTED` --> `PROVIDED_REVIEW`
`ACCEPTED`/`SUBMITTED` --> `REQUESTED_REVIEW`

PR #136, opened 2016-04-04
Upgrade upgrade-google-api-python-client

joshblum:


joshblum:
still want to manually test on staging @marcua what is the best way to do this?

marcua:
lgtm

PR #138, opened 2016-04-05
Update gulpfile to only process scss

joshblum:


orchestra/static/orchestra/common/css/lib/style.css
```diff
@@ -248,27 +248,6 @@ ul.top-menu > li > .logout {
     padding-bottom: 10px;
 }
 
-/*LEFT NAVIGATION ICON*/
```
joshblum, line 251: note: these image files are missing and fail the whitenoise build. the class is unused, @derek-schultz says it's ok to remove

gulpfile.js
```diff
@@ -19,13 +25,10 @@ var files = {
...
     'orchestra/static/orchestra/common/css/orchestra.scss',
-    'orchestra/static/orchestra/common/css/registration.scss',
-    'orchestra/static/orchestra/dashboard/css/dashboard.scss',
-    'orchestra/static/orchestra/task/css/task.scss',
   ],
   all_scss: [],
-  scripts: [
-    '!' + buildDestination + '/**',
+  jslint: [
+    './gulpfile.js', // Lint ourselves!
```
marcua, line 31: :)

marcua:
lgtm, with one concern---a bunch of files were deleted and added, rather than renamed---can we just rename?

joshblum:
handled offline. we good

PR #139, opened 2016-04-05
Create data model for time entry.

lydiagu:


orchestra/admin/__init__.py
```diff
@@ -7,6 +7,7 @@
 from orchestra.models import Step
 from orchestra.models import Task
 from orchestra.models import TaskAssignment
+from orchestra.models import TimeEntry
```
lydiagu, line 8: what is the import convention for orchestra? alphabetical?

jrbotros, line 8: Yep! We break imports into three sections based on their source: stdlib, third-party, and internal (sometimes more for more granularity, e.g., if there's a lot of `django` imports). Each section looks like this:

```
import <module>
from <module> import <lowercase_name>
from <module> import <uppercase_name>
```

orchestra/models.py
```diff
@@ -499,6 +499,30 @@ class Status:
...
+        date (datetime.date):
+            The date the work was done.
+        time_worked (datetime.timedelta):
+            Amount of time worked.
+        assignment (orchestra.models.TaskAssignment):
+            The task assignment for which the work was done.
+        description (str): optional
+            Brief description of the work done.
+    """
+    class Meta:
```
jrbotros: Are we supposed to include these on models? I never have

jrbotros:
lgtm

PR #145, opened 2016-04-07
Suggest way to mock logging instead of just disabling all logs.

lydiagu:


lydiagu:
@marcua @joshblum @jrbotros We can mock logging instead of just turning off all logs (in beanstalk_dispatch.tests.test_dispatcher). I'm not sure what the best way of doing this is, so this is just a suggestion.

marcua:
lgtm, assuming you meant to remove the `setup_models` call

PR #146, opened 2016-04-07
Migrate snapshot data to iterations

jrbotros:


orchestra/tests/helpers/iterations.py
```diff
@@ -5,32 +5,31 @@
 from orchestra.utils.task_properties import get_iteration_history
 
 
-def verify_iterations(test, task_id):
+def verify_iterations(task_id):
```
jrbotros, line 8: Converted this so we can use it outside of tests, not sure what best practices are for this kind of thing?

orchestra/utils/task_properties.py
```diff
@@ -31,11 +32,17 @@ def current_assignment(task):
             The in-progress assignment for `task`.
     """
     assignments = assignment_history(task)
-    if task.status == Task.Status.POST_REVIEW_PROCESSING:
-        # Get second-to-last assignment, since reviewer rejected
-        return assignments.reverse()[1]
-    else:
+    processing = task.assignments.filter(
```
jrbotros, line 35: This function didn't support multiple successive rejections down the review hierarchy.

orchestra/tests/helpers/iterations.py
```diff
@@ -92,14 +90,13 @@ def _verify_iteration_data(test, iterations):
         siblings = assignment.iterations.order_by('start_datetime')
         if iteration.status == Iteration.Status.PROCESSING:
             # Iterations should not have data until submitted
-            test.assertEqual(iteration.submitted_data, {})
+            assert iteration.submitted_data == {}
```
lydiagu, line 93: using assert instead of self.assertEqual won't print out what the expected/actual values are:

```
    assert b == c
AssertionError
```

vs.

```
    self.assertEqual(b, c)
AssertionError: 1 != 2
```

Can we add the functions to the base test case?

orchestra/utils/task_properties.py
lydiagu, line 35: Add a unit test for this case? or is it covered by your existing unit tests?

lydiagu:
migration lgtm, left a couple comments on test update

orchestra/utils/task_properties.py
jrbotros, line 35: Will do this in the next PR!

orchestra/tests/helpers/iterations.py
jrbotros, line 93: discussed offline, leaving for now

PR #162, opened 2016-04-11
Use django rest framework api views for generic APIs

lydiagu:
@marcua @jrbotros what do we think about using rest framework to reduce work for CRUD apis?

### DIFF
--- a/example_project/example_project/orchestra_settings.py
+++ b/example_project/example_project/orchestra_settings.py
@@ -107,7 +107,10 @@ def setup_orchestra(settings_module_name):
 
     # Don't authenticate users without a view explicitly calling for it
     settings.REST_FRAMEWORK = {
-        'DEFAULT_AUTHENTICATION_CLASSES': (),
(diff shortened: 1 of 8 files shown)


example_project/example_project/orchestra_settings.py
```diff
@@ -108,7 +108,10 @@ def setup_orchestra(settings_module_name):
 
     # Don't authenticate users without a view explicitly calling for it
     settings.REST_FRAMEWORK = {
-        'DEFAULT_AUTHENTICATION_CLASSES': (),
+        'DEFAULT_AUTHENTICATION_CLASSES': (
```
marcua, line 110: remind me to think about this

lydiagu, line 110: reminding

I can see us removing BasicAuth - we don't use it and probably shouldn't allow it

marcua:
LGTM assuming the Orchestra API for creating projects and the user interface still loads.

lydiagu:
yup can create projects and user interface is good.

PR #168, opened 2016-04-12
Update timer backend to allow only one timer per worker

lydiagu:


orchestra/tests/helpers/__init__.py
```diff
@@ -73,3 +74,17 @@ def assertModelInstanceExists(self,
     def assertModelInstanceNotExists(self, model_class, lookup_attributes):
         with self.assertRaises(model_class.DoesNotExist):
             model_class.objects.get(**lookup_attributes)
+
+
+class OrchestraTestCase(OrchestraTestHelpersMixin, TestCase):
+    # NOTE(lydia): Mixin should go before TestCase because when mixin calls
```
lydiagu, line 80: @marcua @joshblum good thing we spent a while figuring out multiple inheritance in python! also http://nedbatchelder.com/blog/201210/multiple_inheritance_is_hard.html

lydiagu, line 80: what I find strange is that do people create mixins which must be inherited with another class?

orchestra/utils/time_tracking.py
```diff
@@ -80,59 +81,65 @@ def save_time_entry(worker, task_id, time_entry_data):
...
+    Returns TaskTimer object associated with worker, creates one if does not
+    exist.
+    """
     try:
-        timer = assignment.timer
+        timer = worker.timer
     except TaskTimer.DoesNotExist:
-        timer = TaskTimer.objects.create(worker=worker,
-                                         assignment=assignment)
+        timer = TaskTimer.objects.create(worker=worker)
```
jrbotros: `get_or_create`?

jrbotros:
looks great!

PR #177, opened 2016-04-14
Update TaskAssignment/Iterations to mark as deleted

lydiagu:
@marcua Should have done this instead with Iteration ha

**Update:** I realized that does not work because we order by `start_datetime` in many places.

**Update:** added cascading deletes to DeleteMixin because revert tests were breaking.


orchestra/utils/models.py
```diff
@@ -8,6 +9,27 @@ class DeleteMixin(object):
     Intended to be used with models that have an `is_deleted` field.
     """
     def delete(self, *args, **kwargs):
+        # Implement cascading deletes.
+        # NOTE(lydia): This ignores any other constraints passed to on_delete.
+        fields = self._meta.get_fields(include_hidden=True)
```
lydiagu, line 14: https://docs.djangoproject.com/en/1.9/ref/models/meta/#migrating-from-the-old-api

orchestra/tests/test_views.py
```diff
@@ -131,8 +131,7 @@ def test_time_entry_delete(self):
         self.request_client.delete(
             reverse('orchestra:orchestra:time_entry',
```
marcua, line 132: sanity check: can workers delete other workers' entries?

orchestra/tests/test_views.py
```diff
@@ -131,8 +131,7 @@ def test_time_entry_delete(self):
...
                     kwargs={'pk': time_entry.id}))
-        time_entry.refresh_from_db()
-        self.assertTrue(time_entry.is_deleted)
+        self.assertFalse(TimeEntry.objects.filter(id=time_entry.id).exists())
```
marcua, line 134: how does this work? do we override `objects` somewhere to ignore `is_deleted`?

marcua, line 134: answering my own question: yes, in `BaseModelManager`

orchestra/models.py
```diff
@@ -518,7 +518,6 @@ class TimeEntry(models.Model):
     description = models.CharField(max_length=200, null=True, blank=True)
     timer_start_time = models.DateTimeField(null=True)
     timer_stop_time = models.DateTimeField(null=True)
-    is_deleted = models.BooleanField(default=False)
```
marcua, line 521: delete docstring

marcua:
if the sanity check answer is `no` and you delete and you delete the docstring, lgtm!

orchestra/tests/test_views.py
lydiagu, line 132: they can't because I added the permissions class `isAssociatedWorker`!

PR #178, opened 2016-04-15
Separate task ajax requests into angular service.

lydiagu:


orchestra/static/orchestra/common/js/orchestra.services.js
```diff
@@ -67,3 +67,33 @@ serviceModule.factory('orchestraService', function() {
   };
   return orchestraService;
 });
+
+serviceModule.factory('orchestraTasks', function($http) {
+
+  var newTask = function(taskType) {
```
jrbotros, line 73: At some point we should probably have an API service so we have all our URLs in one place, but totally good for now!

orchestra/static/orchestra/common/js/orchestra.services.js
```diff
@@ -67,3 +67,33 @@ serviceModule.factory('orchestraService', function() {
...
+    for (var taskType in tasks) {
+      if (taskType != 'complete') {
+        numTasks += tasks[taskType].length;
+      }
+    }
+    return numTasks;
+  };
+
+  var service = {
+    newTask: newTask,
```
jrbotros, line 93: Any reason you're not defining these functions inline? Usually our style but I don't have strong feelings.

jrbotros:
lgtm!

orchestra/static/orchestra/common/js/orchestra.services.js
lydiagu, line 93: More readable in general. I was copying the high level structure in https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/static/orchestra/common/js/orchestra.services.js and just now noticed that the nested functions are inline.

orchestra/static/orchestra/common/js/orchestra.services.js
lydiagu, line 73: Yes, agreed!

PR #179, opened 2016-04-18
Add date-range filtering for TimeEntry.

lydiagu:


orchestra/project_api/serializers.py
```diff
@@ -159,10 +159,10 @@ def get_in_progress_task_data(self, obj):
 
 class TimeEntrySerializer(serializers.ModelSerializer):
 
-    worker = serializers.ReadOnlyField(source='worker.id')
-
     class Meta:
         model = TimeEntry
+        fields = ('id', 'date', 'time_worked', 'description', 'assignment')
```
lydiagu, line 164: cleaned up the serializer so we don't allow updating of fields that shouldn't be updated.

orchestra/filters.py
```diff
@@ -0,0 +1,14 @@
+import django_filters
+
+from rest_framework import filters
+
+from orchestra.models import TimeEntry
+
+
+class TimeEntryFilter(filters.FilterSet):
+    min_date= django_filters.DateFilter(name="date", lookup_type='gte')
```
jrbotros: small typo

jrbotros: `min_date = d...`

requirements.txt
```diff
@@ -25,6 +25,7 @@ python-dateutil==2.5.2
 django-hijack==2.1.0
 django-hijack-admin==2.1.0
 django-money==0.7.6
+django-filter==0.13.0
```
jrbotros, line 28: Throw this in alphabetical order

jrbotros:
lgtm

requirements.txt
lydiagu, line 28: i don't think this file is in alphabetical order. should it be? I think generally requirements.txt is append? Changing the order of the requirements can mess up the dependency versions.

jrbotros, line 28: Fair enough, I don't have strong feelings about it!

PR #192, opened 2016-04-21
Round up minutes for all work time displays

jrbotros:


orchestra/static/orchestra/timing/time-entry.js
```diff
@@ -34,6 +35,32 @@
   };
 
   /**
+   * Rounds the duration to the next minute without affecting original data.
+   */
+  moment.duration.fn.roundMinute = function() {
+    var duration = this;
+    return moment.duration({
+      hours: duration.get('hours'),
+      minutes: duration.get('minutes') + 1
```
lydiagu: what if seconds = 0?

lydiagu:
done reviewing

PR #203, opened 2016-04-22
Improve usability across admin views

jrbotros:


orchestra/models.py
```diff
@@ -506,6 +507,9 @@ class TimeEntry(BaseModel):
             Server timestamp for timer stop (not null if TimeEntry is
             created using work timer)
     """
+    class Meta:
+        verbose_name_plural = "time entries"
```
marcua: single quotes

orchestra/admin.py
```diff
@@ -0,0 +1,190 @@
...
+    search_fields = ('workflow__slug', 'slug', 'name', 'description',)
+    list_filter = ('workflow',)
+
+
+def edit_link(instance, text=None):
+    if not instance:
+        return None
+    if text is None:
+        text = str(instance)
+    change_url_name = 'admin:{}_change'.format(instance._meta.db_table)
```
marcua: sanity check---the views are named after the DB tables?  it feels like that breaks encapsulation, but if you've read that this is how it works, then OK

orchestra/admin.py
```diff
@@ -0,0 +1,190 @@
...
+from orchestra.models import Worker
+from orchestra.models import WorkerCertification
+from orchestra.models import Workflow
+from orchestra.models import WorkflowVersion
+
+admin.site.site_header = 'Orchestra'
+admin.site.site_title = 'Orchestra'
+admin.site.index_title = 'Orchestra'
+
+
```
marcua, line 22: sanity check---the number of admin.register reigstrations is the same as it was before this PR?

marcua:
lgtm assuming the two sanity checks pan out

orchestra/admin.py
jrbotros, line 22: ✅

orchestra/admin.py
jrbotros: Agreed—will work with other `_meta` props instead.

orchestra/models.py
jrbotros: ✅

PR #204, opened 2016-04-22
Remove excessive module cruft

jrbotros:


marcua:
sanity check: do you have to update any docs?

marcua:
sanity check: have you run through the journalism example end-to-end?

PR #207, opened 2016-04-23
Simplify task prerequisite data serialization format

jrbotros:
We previously had a nested prerequisite data format (e.g., `vm.taskAssignment.prerequisites.step3.prerequisites.step2.prerequisites.step1...`). This changes the representation to a simple object keyed by step slug, since they're all unique.

### DIFF
--- a/orchestra/utils/task_lifecycle.py
+++ b/orchestra/utils/task_lifecycle.py
@@ -802,7 +802,7 @@ def previously_completed_task_data(task):
     """
     Returns a dict mapping task prerequisites onto their
     latest task assignment information.  The dict is of the form:
-    {'previous-slug': {task_assignment_data}, ...}
+    {'previous-slug': {latest_assignment_data}, ...}
 
     Args:
         task (orchestra.models.Task): The specified task object.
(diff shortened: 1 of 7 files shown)


orchestra/utils/task_lifecycle.py
```diff
@@ -819,27 +819,25 @@ def previously_completed_task_data(task):
...
     """
-    step = task.step
-    prerequisites = {}
+    # Find all prerequisite steps in graph
+    to_visit = list(task.step.creation_depends_on.all())
+    prerequisite_steps = set(to_visit)
+    while to_visit:
+        current_step = to_visit.pop()
+        for step in current_step.creation_depends_on.all():
+            if step not in prerequisite_steps:
```
marcua: given set semantics, no need for this if :)

orchestra/utils/task_lifecycle.py
```diff
@@ -819,27 +819,25 @@ def previously_completed_task_data(task):
...
+                to_visit.append(step)
+                prerequisite_steps.add(step)
+
+    data = {}
```
marcua: prerequisite_data

orchestra/utils/task_lifecycle.py
```diff
@@ -819,27 +819,25 @@ def previously_completed_task_data(task):
...
 
-    for required_step in step.creation_depends_on.all():
-        required_task = Task.objects.get(step=required_step,
-                                         project=task.project)
         if required_task.status != Task.Status.COMPLETE:
             raise TaskDependencyError('Task depenency is not satisfied')
-
-        task_details = get_task_details(required_task.id)
-
         assignment = assignment_history(required_task).last()
```
marcua, line 831: we deleted all of this `task_details` and `task_assignment_details` stuff---any idea what it was for and what we lose/don't lose out on now?

marcua:
sanity check: do you have to update any docs?

---

@jrbotros
✅

marcua:
sanity check: does the journalism example work end-to-end?

orchestra/utils/task_lifecycle.py
jrbotros, line 831: Those methods are our current serializers for tasks and assignments (though I'm looking to replace them with `rest_framework` serializers soon); since we're just returning prerequisite data here, I didn't think we needed the extraneous task/assignment information. But you could imagine a machine task needing the id of a previous task for something? I could be wrong here.

marcua, line 831: if tests pass, i think it's ok---especially since the implementation of the function is now a lot closer to its docstring :)

assuming the sanity checks in my comments hold true, then :shipit:

orchestra/utils/task_lifecycle.py
jrbotros: ✅

PR #211, opened 2016-04-26
Ensure migrations are reviewed and present during builds

joshblum:


orchestra/migrations/0044_auto_20160426_0044.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('orchestra', '0043_tasktimer_description'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
marcua, line 15: wouldn't this not have been caught? :P

marcua:
lgtm, but note my criticism and consider adding another check :)

orchestra/migrations/0044_auto_20160426_0044.py
joshblum, line 15: The missing migration file was caught, we can add another check, is `AlterModelOptions` something we want to be manually reviewed?

PR #213, opened 2016-04-26
Communication preferences

joshblum:
depends on https://github.com/unlimitedlabs/orchestra/pull/211


Makefile
```diff
@@ -11,6 +11,8 @@ lint:
 	flake8 . && gulp lint --production
 
 test: lint
```
marcua, line 13: rebase master before merging?

orchestra/models.py
```diff
@@ -523,3 +525,48 @@ class PayRate(PayRateMixin, models.Model):
...
+
+
+class CommunicationPreference(models.Model):
+    """
+    A CommunicationPreference object defines how a User would like to contacted
+    for a given CommunicationType.
+
+    Attributes:
+        user (django.contrib.auth.models.User):
+            Django user whom the preference represents.
```
marcua: whom -> that ?

orchestra/models.py
```diff
@@ -523,3 +525,48 @@ class PayRate(PayRateMixin, models.Model):
...
+        type (CommunicationType):
+            The type of communication to which this preference applies.
+    """
+
+    class CommunicationMethods(ChoicesEnum):
+        SLACK = 'slack'
+        EMAIL = 'email'
+
+    user = models.ForeignKey(User)
+    methods = MultiSelectField(
```
marcua: how is this different from a bitfield?

which do we want?  i know bitfields better, i guess, but happy to be easily persuaded

orchestra/models.py
```diff
@@ -523,3 +525,48 @@ class PayRate(PayRateMixin, models.Model):
...
+        choices=CommunicationMethods.choices(), blank=True)
+    type = models.ForeignKey(CommunicationType)
```
marcua: communication_type to avoid reserved-ish word

orchestra/models.py
```diff
@@ -523,3 +525,48 @@ class PayRate(PayRateMixin, models.Model):
...
+    Attributes:
+        user (django.contrib.auth.models.User):
+            Django user whom the preference represents.
+        methods (ChoicesEnum):
+            The ways in which the user would like to be contacted.
+        type (CommunicationType):
+            The type of communication to which this preference applies.
+    """
+
+    class CommunicationMethods(ChoicesEnum):
```
marcua: CommunicationMethod

orchestra/models.py
marcua: i feel like this should also be a MultiSelectField/BitField of CommunicationType, where CommunicationType is an enum:
- multi/bit because you can imagine wanting to be alerted in multiple ways
- enum for the same reason methods are enums---we'll want to talk about them programatically. this is assuming I understand the field: it's things like staffbot, etc?

kkamalov: I think having the field as ForeignKey makes sense. We will just have multiple CommunicationMethod objects with different types. You can imagine Users who want to be contacted via email for some things and via slack for other.

orchestra/models.py
kkamalov: ++ for bitfields

Makefile
joshblum, line 13: always :) it's just that the PR was diffed against master and i was waiting on the other review :)

orchestra/models.py
joshblum: done

orchestra/models.py
joshblum: bitfields is better i think, this multiselect package isn't actively maintained anyway

orchestra/models.py
joshblum: done

orchestra/models.py
joshblum: Yeah i originally was going to do a multi/bit field but I think the foreign key makes sense since a user can have difference preferences for each type.

orchestra/models.py
joshblum: actually going to leave as is to be consistent w/product models (i.e. `PurposeFlags`)

orchestra/models.py
marcua: that's not an argument for a foreign key---you can accomplish the same thing with an enum, and we'll want to refer to the preferences programmatically

orchestra/models.py
marcua: The python3 docs suggest it should be singular: https://docs.python.org/3/library/enum.html

orchestra/utils/models.py
```diff
@@ -62,3 +66,98 @@ class BaseModel(DeleteMixin, models.Model):
...
+            'MY_FIRST_ITEM': {
+                'value': 0,
+                'description': 'My first name'
+            },
+            ...
+        }
+        """
+        return {item.name: item.as_dict() for item in cls}
+
+    @classmethod
```
joshblum: note, this is new

joshblum: try to pass value to constructor

joshblum:
@marcua lgtm offline

PR #215, opened 2016-04-26
Delete Time Entries marked for delete so we can make worker_id field …

lydiagu:
…required.


lydiagu:
@marcua I don't love using raw queries, but haven't spent too much time thinking about how to delete the objects since I overrode the model manager

marcua:
Doesn't model freezing remove the methods and managers, so you can use .objects.filter(...).delete()?

PR #216, opened 2016-04-26
Add account settings page for user settings

joshblum:
screen shot 2016-04-26 at 12 52 19
screen shot 2016-04-26 at 12 52 15
screen shot 2016-04-26 at 12 51 21
screen shot 2016-04-26 at 12 56 01


marcua:
If you change your email, does it confirm via email?

marcua:
UX: the active tab should be visually denoted

marcua:
Also add slack username

orchestra/accounts/views.py
```diff
@@ -0,0 +1,87 @@
...
+from registration.models import RegistrationProfile
+from registration.views import RegistrationView
+
+from orchestra.accounts.forms import UserForm
+from orchestra.accounts import signals
+
+UserModel = get_user_model()
+
+
+class OrchestraRegistrationView(RegistrationView):
```
marcua: why do we need this now whereas we didn't need it before?

orchestra/static/orchestra/common/css/orchestra.scss
```diff
@@ -1,6 +1,7 @@
-@import 'orchestra/static/orchestra/common/css/registration.scss';
+@import 'orchestra/static/accounts/scss/registration';
```
marcua, line 1: .scss no longer needed?  why different from the other files here?

orchestra/templates/registration/password_change_done.html
```diff
@@ -1,9 +1,11 @@
-{% extends 'registration/registration_base.html' %}
+{% extends "registration/registration_base.html" %}
```
marcua, line 1: why did we change all of these files?

orchestra/urls.py
```diff
@@ -1,20 +1,19 @@
...
         include('orchestra.api_urls', namespace='orchestra')),
     url(r'^app/?', index, name='index'),
     url(r'^status/', status, name='status'),
-
-    # We have to override the register endpoint to emit the
-    # `orchestra_user_registered` signal
-    url(r'^accounts/register/$',
-        OrchestraRegistrationView.as_view(),
-        name='registration_register'),
+    url(r'^favicon.ico/$', RedirectView.as_view(
```
marcua, line 14: the favicon already worked---why add this?

orchestra/static/orchestra/common/css/orchestra.scss
joshblum, line 1: according to @derek-schultz the proper way is to only have `_filename.scss` and then the scss import works as i have it. we haven't followed this convention in orchestra but should going forward

orchestra/templates/registration/password_change_done.html
joshblum, line 1: to get the tabs in place

orchestra/urls.py
joshblum, line 14: there is always a 404 since the browser checks for the url `favicon.ico`, just made this change in product so thought it would be good to propagate

joshblum:
slack username will come separately. this PR is big enough as is

joshblum:
no email confirmation rn, note we also don't do this on product currently. should we add it to both?

joshblum:
screen shot 2016-04-26 at 15 04 28

PR #221, opened 2016-04-27
Add PhoneNumberField to Worker

joshblum:


orchestra/models/core/models.py
```diff
@@ -184,10 +185,13 @@ class Worker(WorkerMixin, models.Model):
             The time the worker was created.
         slack_username (str):
             The worker's Slack username if Slack integration is enabled.
+        phone (str):
```
kkamalov, line 188: is it str?

joshblum, line 188: yeah, the `PhoneNumberField` is stored internally as a `CharField`

PR #231, opened 2016-04-28
Refactor code to bring staffing logic to StaffBot

joshblum:

PR #232, opened 2016-04-29
Staffbot parsing

joshblum:
depends on #231

### DIFF
--- a/orchestra/bots/staffbot.py
+++ b/orchestra/bots/staffbot.py
@@ -32,9 +32,9 @@ class Bot(object):
                  allowed_user_names=None,
                  allowed_commands=None):
         """
-            Base configuration for the bot to validate messages. If a variable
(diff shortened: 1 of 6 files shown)


orchestra/bots/staffbot.py
```diff
@@ -63,25 +76,83 @@ def validate(self, data):
                             fieldname, value))
         return data
 
+    def format_slack_message(self, text,
+                             attachments=None,
+                             response_type='ephemeral'):
+        """
```
marcua, line 82: better to use this docstring format (no tab beyond triple quotes, and either backticks on variable names or put them in an attribute list:

https://github.com/unlimitedlabs/orchestra/blob/master/orchestra/models/core/models.py#L54

marcua:
lgtm except for docstring, fine to skip if you're pressed for time.

PR #234, opened 2016-04-29
Added worker slack_ids. Datamigration to port ids for existing users

kkamalov:


orchestra/accounts/forms.py
```diff
@@ -28,7 +30,26 @@ class WorkerForm(forms.ModelForm):
...
+        if slack_username:
+            slack = Slacker(settings.SLACK_EXPERTS_API_KEY)
+            slack_id = slack.users.get_user_id(slack_username)
+            if slack_id is None:
+                raise forms.ValidationError(
+                    'Incorrect Slack Username Provided')
+            self.cleaned_data['slack_id'] = slack_id
+        return self.cleaned_data['slack_username']
+
+    def save(self):
```
joshblum: I think you should model the save method off of https://github.com/unlimitedlabs/crowdsurfing/blob/master/product/accounts/forms.py#L57 with the `commit` flag

orchestra/accounts/forms.py
```diff
@@ -28,7 +30,26 @@ class WorkerForm(forms.ModelForm):
...
+        instance = super().save()
+        if hasattr(self.cleaned_data, 'slack_id'):
```
joshblum: `instance.slack_id = self.cleaned_data.get('slack_id)`

orchestra/accounts/forms.py
```diff
@@ -28,7 +30,26 @@ class WorkerForm(forms.ModelForm):
...
+        exclude = ('user', 'start_datetime', 'slack_id')
+
+    def clean_slack_username(self):
+        slack_username = self.cleaned_data.get('slack_username')
+        if slack_username:
+            slack = Slacker(settings.SLACK_EXPERTS_API_KEY)
+            slack_id = slack.users.get_user_id(slack_username)
+            if slack_id is None:
+                raise forms.ValidationError(
+                    'Incorrect Slack Username Provided')
```
joshblum: No caps on the error message

orchestra/migrations/0053_worker_slack_id.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('orchestra', '0052_auto_20160428_2142'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='worker',
+            name='slack_id',
```
joshblum: `slack_user_id` instead? the slash command returns a field called `user_id` so it might be better to be consistent with this (and also there are other types of ids like channels)

orchestra/migrations/0054_add_slack_ids.py
```diff
@@ -0,0 +1,61 @@
...
+logger = logging.getLogger(__name__)
+
+def add_worker_slack_ids(apps, schema_editor):
+    Worker = apps.get_model('orchestra', 'Worker')
+    if not hasattr(settings, 'SLACK_EXPERTS_API_KEY'):
+        return
+
+    slack = Slacker(settings.SLACK_EXPERTS_API_KEY)
+    slack_response = slack.users.list()
+    if not slack_response.successful:
```
joshblum: add logger warning here?

orchestra/migrations/0054_add_slack_ids.py
```diff
@@ -0,0 +1,61 @@
...
+from django.db import migrations
+from django.conf import settings
+from slacker import Slacker
+
+import logging
+logger = logging.getLogger(__name__)
+
+def add_worker_slack_ids(apps, schema_editor):
+    Worker = apps.get_model('orchestra', 'Worker')
+    if not hasattr(settings, 'SLACK_EXPERTS_API_KEY'):
```
joshblum: logger warning that the migration exited

orchestra/migrations/0054_add_slack_ids.py
```diff
@@ -0,0 +1,61 @@
...
+    members = slack_response.body.get('members')
+    for member in members:
+        if member.get('is_bot'):
+            continue
+        worker = get_object_or_None(Worker,
+                                    slack_username=member['name'])
+        if worker is None:
+            # Try to match user by first name and last name.
+
+            if (not hasattr(member['profile'], 'first_name') or
```
joshblum: use `getattr` here with a default of `None`

orchestra/migrations/0054_add_slack_ids.py
joshblum: check that the key is valid i.e. not an empty string or something, I think this is causing the migration to hang

orchestra/accounts/forms.py
```diff
@@ -28,7 +30,26 @@ class WorkerForm(forms.ModelForm):
     class Meta:
         model = Worker
         fields = ('slack_username', 'phone')
-        exclude = ('user', 'start_datetime')
+        exclude = ('user', 'start_datetime', 'slack_id')
+
+    def clean_slack_username(self):
+        slack_username = self.cleaned_data.get('slack_username')
+        if slack_username:
+            slack = Slacker(settings.SLACK_EXPERTS_API_KEY)
```
joshblum: need to mock this out in testing now

marcua: i feel like slack implementation stuff should live in a `slack` module that we call, without having to pass in an api key, etc.

orchestra/accounts/forms.py
kkamalov: k

orchestra/accounts/forms.py
kkamalov: k

orchestra/accounts/forms.py
kkamalov: k

orchestra/accounts/forms.py
kkamalov: k

orchestra/migrations/0053_worker_slack_id.py
kkamalov: k

orchestra/scripts/add_slack_user_ids.py
```diff
@@ -0,0 +1,49 @@
...
+from slacker import Slacker
+
+from orchestra.models import Worker
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def add_worker_slack_ids():
+    if not hasattr(settings, 'SLACK_EXPERTS_API_KEY'):
```
joshblum: `getattr(settings, 'SLACK_EXPERTS_API_KEY`, None)`if the key is`''`, the request will fail so it would be better to catch here

orchestra/utils/slack.py
```diff
@@ -0,0 +1,8 @@
+from django.conf import settings
+from slacker import Slacker
+
+
+def get_slack_user_id(slack_username):
+    slack = Slacker(settings.SLACK_EXPERTS_API_KEY)
```
joshblum: let's move this object into slack as well just using the `settings.SLACK_EXPERTS_API_KEY` as default

orchestra/scripts/add_slack_user_ids.py
```diff
@@ -0,0 +1,49 @@
+from annoying.functions import get_object_or_None
```
marcua, line 1: i assume this script won't run b/c django isn't configured?  make it a management command?

kkamalov, line 1: yes. it does not do anything right now

joshblum, line 1: would it be better as a migration actually that is skipped during testing? otherwise we need more setup docs about running this script etc

orchestra/scripts/add_slack_user_ids.py
kkamalov: done

PR #237, opened 2016-04-29
Fix rendering of saved options

joshblum:
screen shot 2016-04-29 at 12 45 27

Options are stacked now (and fixes a bug where they would be unchecked)


kkamalov:
yeah, wanted to tell you that display was not working but it was saving :)

PR #244, opened 2016-05-02
Add more tests for staffing related code

joshblum:


orchestra/bots/tests/test_basebot.py
```diff
@@ -30,6 +28,33 @@ def test_request_validation(self):
...
+        text = 'test_command 5'
+        mock_slack_data['text'] = text
+        result = bot.dispatch(mock_slack_data)
+        self.assertEqual(result, text)
+
+        # Test invalid command
+        text = 'invalid'
+        mock_slack_data['text'] = text
+        result = bot.dispatch(mock_slack_data)
+        self.assertEqual(bot.no_command_found(text), result)
```
kkamalov, line 66: test the case where the command is correct but the param is not provided

PR #248, opened 2016-05-03
Specify assignment policy as a python path

joshblum:


orchestra/assignment_policies/simple.py
```diff
@@ -0,0 +1,61 @@
+from orchestra.core.errors import AssignmentPolicyError
+from orchestra.models import Task
+from orchestra.utils.task_lifecycle import assign_task
+from orchestra.utils.task_properties import assignment_history
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def anyone_certified(task, related_steps):
```
marcua: no related steps?

orchestra/assignment_policies/simple.py
```diff
@@ -0,0 +1,61 @@
+from orchestra.core.errors import AssignmentPolicyError
```
marcua: note sure if this needs to be in a module called simple :)

orchestra/assignment_policies/simple.py
```diff
@@ -0,0 +1,61 @@
...
+        orchestra.core.errors.AssignmentPolicyError:
+            Machine steps cannot be included in an assignment policy.
+    """
+    if related_steps is None:
+        raise AssignmentPolicyError('No related steps given')
+
+    workflow_version = task.step.workflow_version
+    for step_slug in related_steps:
+        step = workflow_version.steps.get(slug=step_slug)
+        if not step.is_human:
```
marcua: is this good to check across all policies instead of this one?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -498,7 +498,8 @@ def test_malformed_assignment_policy(self):
             slug='machine_step',
             is_human=False,
             assignment_policy={
-                'policy': 'previously_completed_steps',
+                'policy': ('orchestra.assignment_policies.'
+                           'simple.previously_completed_steps'),
                 'steps': ['step_0']
```
marcua: kwargs --- not every policy requires this

orchestra/assignment_policies/simple.py
```diff
@@ -0,0 +1,61 @@
...
+    related_tasks = (
+        Task.objects
+        .filter(step__slug__in=related_steps, project=task.project)
+        .select_related('step'))
+    for related_task in related_tasks:
+        entry_level_assignment = assignment_history(related_task).first()
+        if entry_level_assignment and entry_level_assignment.worker:
+            try:
+                return assign_task(entry_level_assignment.worker.id, task.id)
+            except Exception:
```
marcua: Should we catch the specific type of exception (WorkerCertification or something like that)?

orchestra/tests/test_task_lifecycle.py
```diff
@@ -498,7 +498,8 @@ def test_malformed_assignment_policy(self):
             slug='machine_step',
             is_human=False,
             assignment_policy={
-                'policy': 'previously_completed_steps',
+                'policy': ('orchestra.assignment_policies.'
```
marcua: this way of describing functions is different from the way we describe machine functions.  we should keep them consistent, though i don't have an opinion which one changes

orchestra/assignment_policies/simple.py
joshblum: why not? shouldn't all the functions should have the same signature?

orchestra/assignment_policies/simple.py
joshblum: should it just be `orchestra/assignment_policies.py`?

orchestra/tests/test_task_lifecycle.py
joshblum: what is this?

i'll add kwargs though, was going to add it in a later PR but I think it makes the most sense now

orchestra/assignment_policies/simple.py
joshblum: This code was just being refactored, I'm fine changing it should we catch multiple things or is `WorkerCertification` enough?

joshblum: (Plus maybe leaving the `Exception` case in as well)

orchestra/tests/test_task_lifecycle.py
joshblum: meaning we should have a `module` key instead of `policy`?

orchestra/assignment_policies/simple.py
marcua: seems fine for now until it balloons in size :)

orchestra/assignment_policies/simple.py
marcua: nope---in our previous discussion, we had a `kwargs` parameter

orchestra/assignment_policies/simple.py
marcua: maybe do WorkerCertification and then Exception.  or leave as-is, since this is just a refactor :)

orchestra/tests/test_task_lifecycle.py
marcua: it would be:

```
"policy_function": {
            "module": "orchestra.assignment_policies.",
            "name": "previously_completed_steps"
        }
```

To mimick this:
 https://github.com/unlimitedlabs/orchestra/blob/assignment-policy-as-function/journalism_workflow/v1/version.json#L95

I actually like your implementation better, and pydoc.locate seems cleaner.  if you opt for your current version, consider 1) naming it `policy_function` and 2) adding a TODO to make machine functions use the same schema

orchestra/assignment_policies/simple.py
joshblum: ah i see i thought `kwargs` was for more specific params but `related_steps` was required by all (except the allow everyone) :)

orchestra/assignment_policies/simple.py
joshblum: we can change it I think it's a better way to handle it :)

orchestra/tests/test_task_lifecycle.py
joshblum: cool i'll do the rename and the todo :)

orchestra/assignment_policies/simple.py
joshblum: done

orchestra/assignment_policies/simple.py
joshblum: done

orchestra/assignment_policies/simple.py
joshblum: done

orchestra/utils/task_lifecycle.py
```diff
@@ -901,64 +901,22 @@ def _preassign_workers(task):
...
                                         'assignment policy.')
-    elif policy == 'previously_completed_steps' and related_steps is not None:
-        task = _assign_worker_from_previously_completed_steps(task,
-                                                              related_steps)
-    elif policy == 'anyone_certified':
-        # Leave the task in the awaiting processing pool
-        pass
-    else:
+    elif not policy:
         raise AssignmentPolicyError('Assignment policy incorrectly specified.')
```
kkamalov, line 913: do we test this errors?

joshblum, line 913: yup in `test_task_lifecycle.py`

kkamalov:
k lgtm

PR #254, opened 2016-05-03
Specify a reviewer and entry_level assignment policy

joshblum:


orchestra/utils/task_lifecycle.py
```diff
@@ -795,7 +795,10 @@ def submit_task(task_id, task_data, iteration_status, worker):
     task.status = next_status
     task.save()
 
-    if task.status == Task.Status.REVIEWING:
+    if task.status == Task.Status.PENDING_REVIEW:
+        # Check the assignment policy to try to assign a reviewer automatically
+        task = _preassign_workers(task, policy_type='reviewer')
```
marcua: it'd be nice to externalize these string somewhere, like POLICY_TYPE_REVIEWER or something

orchestra/utils/task_lifecycle.py
```diff
@@ -896,14 +899,17 @@ def end_project(project_id):
         notify_status_change(task, assignment_history(task))
 
 
-def _preassign_workers(task):
+def _preassign_workers(task, policy_type='entry_level'):
```
marcua: i think we should make this arg explicit

orchestra/utils/task_lifecycle.py
joshblum: I had it in an enum but then none of the other json keys were so i took it out

orchestra/utils/task_lifecycle.py
joshblum: ok let's do it and i'll add the enum back

PR #255, opened 2016-05-04
Add task assignment for staffing winner.

kkamalov:


orchestra/communication/tests/test_staffing.py
```diff
@@ -36,6 +40,16 @@ def test_handle_staffing_response_is_available(self):
         self.assertTrue(response.is_winner)
         self.assertEqual(StaffingResponse.objects.all().count(), old_count + 1)
 
+        task_assignment = (TaskAssignment.objects
```
joshblum: i don't understand what this test does, shouldn't you check the state of the object before and after the request is made?

orchestra/models/core/model_mixins.py
```diff
@@ -92,6 +92,20 @@ def __str__(self):
 
 class TaskMixin(object):
 
+    def is_worker_assigned_to_task(self, worker):
```
joshblum: rename `is_worker_assigned` since we're calling it from a `Task` object already

orchestra/utils/task_properties.py
```diff
@@ -70,22 +70,3 @@ def all_workers(task):
             A list of all workers involved with `task`.
     """
     return [assignment.worker for assignment in assignment_history(task).all()]
-
```
joshblum, line 73: we should move these functions over the mixins as well, just add a TODO and we'll do it in another PR

PR #261, opened 2016-05-04
Only superusers and CE can send slack messages to StaffBot.

kkamalov:


orchestra/bots/staffbot.py
```diff
@@ -214,3 +217,20 @@ def _send_staffing_request_by_slack(self, staffing_request, message):
...
+        """
+        Handle a request received from slack. First we validate the
+        request and then pass the message to the appropriate handler.
+        """
+        slack_user_id = data.get('user_id')
+        username = data.get('user_name')
+
+        worker = get_object_or_None(Worker, slack_user_id=slack_user_id)
+        if worker is None:
+            raise SlackUserUnauthorized('Worker {} not found'.format(username))
```
joshblum, line 239: We should return a json response instead of raising. Otherwise the end user will just get the HTML for a 500 page back in slack

joshblum, line 239: (or we can catch this in the view and format there?)

orchestra/bots/staffbot.py
```diff
@@ -214,3 +217,20 @@ def _send_staffing_request_by_slack(self, staffing_request, message):
...
+        elif (not is_project_admin(worker.user) and
```
joshblum: let's edit the `is_project_admin` to check for superusers as well

orchestra/bots/tests/test_views.py
```diff
@@ -17,7 +17,12 @@ class StaffBotViewTest(OrchestraTestCase):
     def setUp(self):
         super().setUp()
         setup_models(self)
-        self.request_client = RequestClient()
+        self.worker = self.workers[0]
+        self.worker.user.is_superuser = True
```
joshblum: we should test with someone who is in the client admin group as well

orchestra/tests/helpers/fixtures.py
```diff
@@ -443,7 +443,8 @@ def _setup_workers(test_case, workers):
                             first_name='test_first_{}'.format(user_id),
                             last_name='test_last_{}'.format(user_id),
                             email='test_user_{}@test.com'.format(user_id)))
-        worker = WorkerFactory(user=user)
+        worker = WorkerFactory(user=user,
+                               slack_user_id='test_user_{}'.format(user_id))
```
joshblum, line 454: why not make this a default in the `WorkerFactory`?

orchestra/bots/tests/test_views.py
```diff
@@ -17,7 +17,12 @@ class StaffBotViewTest(OrchestraTestCase):
...
+        self.worker.user.save()
```
joshblum, line 22: add a test to verify that an unauthorized user receives a json error message (status code should be 200 so slack can display the message nicely though)

orchestra/bots/staffbot.py
kkamalov, line 239: i forgot to add a views.py that catches it

orchestra/tests/helpers/fixtures.py
kkamalov, line 454: because we only set test user information in the line above

joshblum:
lgtm

PR #264, opened 2016-05-05
Add assign_worker_task API endpoint

joshblum:


orchestra/tests/test_project_api.py
```diff
@@ -263,6 +264,44 @@ def test_workflow_types(self):
...
+        self.assertEquals(response.status_code, 200)
+        data = load_encoded_json(response.content)
+        self.assertEqual(data['success'], success)
+        return data
+
+    def test_assign_worker_to_task(self):
+        worker = self.workers[0]
+
+        # We should be able to assign to an unassigned task
+        task = self.tasks['awaiting_processing']
```
kkamalov, line 285: can you provide invalid arguments? what happens then. also maybe test if worker does not exist

PR #270, opened 2016-05-06
Settings UX

marcua:
Doing the easy parts of @meredithblumenstock's feedback


marcua:
Pending a bugfix we reported (https://github.com/stefanfoulis/django-phonenumber-field/issues/125), this should be good to go:

image

image

image

orchestra/static/accounts/scss/_registration.scss
```diff
@@ -4,6 +4,9 @@
 /* Login/account-related pages */
 .login-wrap {
   font-size: 1.3rem;
+  .account-tabs selected {
```
joshblum: rm? or is this used to clear the previous css?

joshblum:
one question but otherwise lgtm

orchestra/static/accounts/scss/_registration.scss
marcua: good catch fixed!

meredithblumenstock:
1)  Add more padding between the checkboxes and the checkbox labels
2) Maybe add more padding between the country code dropdown and the phone number text field (low priority)
3) Can you change the card color to match the white shade of the tab?

PR #302, opened 2016-05-12
Update StaffBot docs to have usage examples

joshblum:


marcua:
lgtm after a few modifications we discussed offline

PR #312, opened 2016-05-18
Staffbot usability updates.

kkamalov:
screen shot 2016-05-18 at 2 33 43 pm

### DIFF
--- a/orchestra/bots/staffbot.py
+++ b/orchestra/bots/staffbot.py
@@ -1,4 +1,5 @@
 from annoying.functions import get_object_or_None
+from django.db.models import Q
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.template import Context
@@ -127,12 +128,13 @@ class StaffBot(BaseBot):
         command = 'restaff {} {}'.format(task_id, username)
         try:
             error_msg = None
-            worker = get_object_or_None(Worker,
-                                        user__username=username)
-            if worker is None:
-                worker = get_object_or_None(Worker,
-                                            slack_username=username)
-            if worker is None:
+
+            worker = Worker.objects.filter(
+                Q(user__username=username) | Q(slack_username=username))
+
+            if worker.exists():
+                worker = worker.first()
+            else:
                 error_msg = self.worker_does_not_exist.format(username)
                 return format_slack_message(
                     command,
--- a/orchestra/bots/staffbot.py
+++ b/orchestra/bots/staffbot.py
@@ -35,8 +35,9 @@ VALID_RESPONSE_TYPES = {'ephemeral', 'in_channel'}
 class StaffBot(BaseBot):
 
     commands = (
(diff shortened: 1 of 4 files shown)


orchestra/bots/staffbot.py
```diff
@@ -110,17 +124,28 @@ def restaff(self, task_id, username,
         The current user for the given username is removed, and a new user
         is found.
         """
-        # TODO(kkamalov): maybe username could also be slack_username
+        command = 'restaff {} {}'.format(task_id, username)
         try:
-            worker = Worker.objects.get(user__username=username)
```
joshblum, line 115: use `Q` operators and just do a single or query so we don't hit the DB twice :)

orchestra/bots/tests/test_basebot.py
```diff
@@ -44,7 +44,7 @@ def test_dispatch(self):
             (r'test_command (?P<test_param>[0-9]+)', 'test_command'),
         )
 
-        def test_command(test_param):
+        def test_command(test_param, command=None):
```
joshblum: what is this for?

orchestra/bots/staffbot.py
```diff
@@ -101,7 +109,13 @@ def staff(self, task_id,
 
         slack_message = self.staffing_success.format(task_id)
         message_experts_slack_group(task.project.slack_group_id, slack_message)
-        return format_slack_message(slack_message)
+        return format_slack_message(
+            command,
+            attachments=[{
+                'color': 'good',
```
joshblum, line 116: can we abstract this to a success/error attachement message so we don't have to repeat? or maybe it's not worth it?

joshblum:
love the new slack messaging look :)

orchestra/bots/staffbot.py
kkamalov, line 116: i dont think it's worth it :P

orchestra/bots/tests/test_basebot.py
```diff
@@ -44,7 +44,7 @@ def test_dispatch(self):
...
             return text
```
joshblum: also how does this function work haha what is `text` should that be `test_param`?

kkamalov: lol. no idea? did you write the code ;P?

kkamalov: text is actually defined couple of lines below

orchestra/bots/tests/test_basebot.py
```diff
@@ -44,7 +44,7 @@ def test_dispatch(self):
             (r'test_command (?P<test_param>[0-9]+)', 'test_command'),
         )
 
-        def test_command(test_param, command=None):
+        def test_command(test_param):
             return text
```
joshblum: should this return 'test_param'?

joshblum:
lgtm after tests

PR #315, opened 2016-05-20
Add project short description to StaffBot requests

marcua:


orchestra/bots/tests/test_staffbot.py
```diff
@@ -268,6 +270,7 @@ def _task_factory(status, path):
 A new task is available for you to work on, if you'd like!  Here are the details:
 
 Project type: the workflow
+Project short description: the coolest project
```
joshblum: Should we just call this description so if a user wants to add more of a description via the slack command we call that short description?

joshblum:
one question, otherwise lgtm

marcua:
Perfect!  Removed `short` after offline discussion.

PR #336, opened 2016-07-28
Option to enable/disable new task buttons

marcua:


marcua:
(received an offline LGTM from @joshblum)

PR #348, opened 2016-08-09
Speedup test suite by using weak hash algo

vhf:
Default password hasher is slow by design, which in turn usually slows down test user creation quite a bit. Relevant doc: https://docs.djangoproject.com/en/1.9/topics/testing/overview/#password-hashing

On my laptop, repeated `time make test` goes from >110s to <60s.

### DIFF
--- a/example_project/example_project/settings.py
+++ b/example_project/example_project/settings.py
@@ -130,4 +130,9 @@ USE_TZ = True
 STATIC_URL = '/static/'
 STATIC_ROOT = 'staticfiles'
 
+
+TESTING = sys.argv[1:2] == ['test']
+if TESTING:
+    PASSWORD_HASHERS = ('django.contrib.auth.hashers.MD5PasswordHasher',)
+
 setup_orchestra(__name__)


example_project/example_project/settings.py
```diff
@@ -130,4 +130,9 @@
 STATIC_URL = '/static/'
 STATIC_ROOT = 'staticfiles'
 
+
+TESTING = sys.argv[1:2] == ['test']
```
derek-schultz, line 134: Why a slice here and not just look at `sys.argv[1]`?

joshblum, line 134: i think because he doesn't check the length, a slice will always return an empty array

vhf:
Unfortunately CI fails, seemingly because it cannot push coverage report.

joshblum:
@vhf yeah it's a bug with coveralls :/, just submitted https://github.com/coagulant/coveralls-python/issues/115 to ask about it. We can review and manually test for the time being, thanks for the contributions!!

PR #349, opened 2016-08-10
Fix a few StaffBot validation bugs

marcua:


orchestra/bots/staffbot.py
```diff
@@ -311,12 +311,15 @@ def validate(self, data):
         slack_user_id = data.get('user_id')
         username = data.get('user_name')
 
-        worker = get_object_or_None(Worker, slack_user_id=slack_user_id)
-        if worker is None:
```
joshblum, line 315: just do `worker = workers.first()`

joshblum:
one small nit and then lgtm!

PR #351, opened 2016-08-10
Limit number of parallel processes to 6

joshblum:
Before:

```
Ran 191 tests in 78.040s
```

After with 6 processes:

```
Ran 191 tests in 80.788s
```

After with 4 processes:

```
Ran 191 tests in 77.101s
```

Small change, but useful as the number of migrations grows


jrbotros:
lgtm

PR #360, opened 2016-10-20
Increase the number of recently closed tasks while limiting time tracking to only recently closed tasks

marcua:
This is a hacky short-term fix in lieu of a more involved overhaul.  It addresses several issues:
- The timesheet report displays tasks that are legitimately assigned to a task assignment as unassigned once the task is no longer in a worker's recently completed tasks.  To bandaid this, I increased the number of recently completed tasks we send to the frontend from 20 to 200.  I tested this on a user with ~100 tasks and saw no serious browser performance degradation on a 2011-era laptop.
- When assigning a new timesheet entry, users were frustrated that all tasks displayed rather than the tasks they are actively working on.  A simple solution would be to only show active tasks, but this would cause an issue if users just closed a project out and then wanted to submit their timesheets.  As a happy medium, this implementation allows users to select from all active tasks and the most recent 5 closed tasks.

A larger overhaul that addresses both of these issues should follow!:)

### DIFF
--- a/orchestra/static/orchestra/common/js/orchestra.services.js
+++ b/orchestra/static/orchestra/common/js/orchestra.services.js
@@ -110,17 +110,23 @@
         }
         return tasks;
       },
-      tasksForType: function(taskType) {
-        return this.tasks[taskType];
-      },
-      numActiveTasks: function() {
-        var numTasks = 0;
-        for (var taskType in this.tasks) {
-          if (taskType != 'complete') {
-            numTasks += this.tasksForType(taskType).length;
+      activeTasks: function() {
+        var tasks = [];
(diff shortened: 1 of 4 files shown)


orchestra/static/orchestra/common/js/orchestra.services.js
```diff
@@ -110,17 +110,23 @@
         }
         return tasks;
       },
-      tasksForType: function(taskType) {
-        return this.tasks[taskType];
-      },
-      numActiveTasks: function() {
```
marcua, line 116: this was only used by the dashboard

jrbotros (approved):
Looks great for now!

PR #361, opened 2016-11-04
Put schedule machine task in an on_commit so task exists

lydiagu:


marcua (approved):
LGTM barring:
- [ ] Tests fail
- [ ] Make sure that the view calling this function is in a `transaction.atomic`

PR #363, opened 2016-11-16
Allow conditional execution of a step via creation policies

joshblum:


orchestra/bots/assignment_policies.py
```diff
@@ -6,3 +6,4 @@ def staffbot_autoassign(task, **kwargs):
     request_cause = StaffBotRequest.RequestCause.AUTOSTAFF.value
     bot = StaffBot()
     bot.staff(task.id, request_cause=request_cause)
+    return task
```
marcua, line 9: why?

orchestra/utils/task_lifecycle.py
```diff
@@ -924,6 +926,35 @@ def end_project(project_id):
         notify_status_change(task, assignment_history(task))
 
 
+def _check_creation_policy(step, project):
+    """
+    Check the creation policy of the given step allows the task to be completed
```
marcua: Check the creation policy of the given step, which will determine whether to create the task.

orchestra/utils/task_lifecycle.py
```diff
@@ -924,6 +926,35 @@ def end_project(project_id):
...
+    Returns:
+        success (boolean) if the task should be created.
+    Raises:
+        orchestra.core.errors.CreationPolicyError:
+            The specified creation policy type is not supported or the policy
+            failed during execution.
+    """
+    policy = step.creation_policy.get('policy_function', {})
+    policy_path = policy.get('path', '')
+    policy_kwargs = policy.get('kwargs', {})
```
marcua, line 946: don't want to default to yes?  makes it easier

orchestra/utils/task_lifecycle.py
```diff
@@ -924,6 +926,35 @@ def end_project(project_id):
...
+    try:
+        policy_function = locate(policy_path)
+        prerequisite_data = get_previously_completed_task_data(step, project)
+        project_data = project.project_data
+        success = policy_function(
```
marcua: return policy_function

orchestra/utils/tests/test_creation_policies.py
```diff
@@ -0,0 +1,41 @@
...
+        # Submit task; next task should not be created, it never is
+        with patch('orchestra.utils.task_lifecycle._is_review_needed',
+                   return_value=False):
+            initial_task = submit_task(initial_task.id, {},
+                                       Iteration.Status.REQUESTED_REVIEW,
+                                       self.workers[0])
+
+        self.assertEqual(project.tasks.count(), 1)
+
+        # Reset project
```
marcua: why do we need to delete?

orchestra/utils/tests/test_task_lifecycle.py
```diff
@@ -612,6 +615,38 @@ def test_malformed_assignment_policy(self):
...
+        initial_task = assign_task(self.workers[4].id,
+                                   project.tasks.first().id)
+        initial_task.status = Task.Status.COMPLETE
+        initial_task.save()
+
+        # Cannot have an invalid blob for the creation_policy
+        with self.assertRaises(CreationPolicyError):
+            create_subsequent_tasks(project)
+
+        # Reset project
```
marcua: necessary?

orchestra/utils/task_lifecycle.py
marcua: oh nvm---we so `get_default` on load if it doesn't exist, so it's alright?

orchestra/bots/assignment_policies.py
joshblum: we don't currently use the value, but policy functions should return the task

orchestra/utils/task_lifecycle.py
joshblum: done

orchestra/utils/task_lifecycle.py
joshblum: yeah. if you create a step manually and don't put the default it will err. we can change this behavior but i was modeling against how we implemented assignment_polices

orchestra/utils/task_lifecycle.py
joshblum: done

orchestra/utils/tests/test_creation_policies.py
joshblum: removed

orchestra/utils/tests/test_task_lifecycle.py
joshblum: nope :)

orchestra/bots/assignment_policies.py
joshblum: https://github.com/b12io/orchestra/blob/master/orchestra/utils/task_lifecycle.py#L962

marcua: ok cool---whereas preassigning workers modifies the task, i don't think this function should.  returning the task suggests it might have been mutated.

that said i don't feel terribly strongly, so if you feel consistency is better, i can understand that.

marcua (approved):
lgtm!

orchestra/bots/assignment_policies.py
joshblum: this is the preassign function, just caught the bug :)

PR #371, opened 2016-12-22
Webpack common component code

jrbotros:


orchestra/static/orchestra/common/components/quill/quill.directive.es6.js
```diff
@@ -0,0 +1,239 @@
+import template from './quill.html'
+
+// TODO(jrbotros): update to v1.0.0
+import Quill from 'quill'
+import 'quill/dist/quill.base.css'
```
marcua, line 5: why the dist one and not the scss file?

jrbotros: This is Quill's internal css, but I accidentally also deleted our additional custom styles 😳 Added back in!

marcua:
lgtm assuming the rest of the 3rd-party stuff is also included if need be

PR #375, opened 2016-12-23
Webpack project management code

jrbotros:


orchestra/static/orchestra/main.es6.js
```diff
@@ -6,6 +6,7 @@ import {
 
 import common from 'orchestra/common/common.module.es6.js'
 import dashboard from 'orchestra/dashboard/dashboard.module.es6.js'
+import projectManagement from 'orchestra/project-management/project-management.module.es6.js'
```
marcua, line 9: Style q: should this just be module.es6.js since it's already in a project management directory?

orchestra/static/orchestra/project-management/js/crosshair.es6.js
```diff
@@ -0,0 +1,64 @@
+export default function crosshair (visUtils, axis) {
```
marcua: Do we need a js  here?

marcua (approved):
LGTM but two organizational questions

orchestra/static/orchestra/main.es6.js
jrbotros: I don't feel strongly either way, just sticking to the convention I started—happy to change them all if you'd like!

orchestra/static/orchestra/project-management/js/crosshair.es6.js
jrbotros: I was hesitant to move things around since this code is more complicated, but now that it's all working again I'll flatten the structure.

PR #383, opened 2017-03-12
Use Django's native JSONField

jrbotros:
Addresses #321, removing our dependency on `django-jsonfield` in favor of Django's internal implementation.


orchestra/models/core/models.py
```diff
@@ -44,7 +44,7 @@ class Workflow(WorkflowMixin, models.Model):
     name = models.CharField(max_length=200)
     description = models.TextField()
     code_directory = models.CharField(max_length=255, unique=True)
-    sample_data_load_function = JSONField(default={})
+    sample_data_load_function = JSONField(default=dict)
```
jrbotros, line 47: See warning: https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#jsonfield

joshblum:
I think you have to migrate the fields right or we would lose data? Or do the fields transfer over easily?

PR #395, opened 2017-05-10
Only include beanstalk urls on worker machines

joshblum:


marcua (changes requested):
I think this isn't necessary in light of https://github.com/b12io/orchestra/blob/3e28d5f2496b77cc2cd8b190b98b354ca8686a77/example_project/example_project/orchestra_settings.py#L144

PR #400, opened 2017-05-15
chore(build): upgrade webpack to version 2

jrbotros:


.circleci/config.yml
```diff
@@ -8,10 +8,14 @@ jobs:
       - checkout
       - run:
           name: Install dependencies
+          environment:
+            PATH: $HOME/.yarn/bin:$PATH
```
joshblum: do we need the path export here and line 16?

jrbotros: Nope, that's making the build break because circle won't let you update env vars like that—removed!

joshblum (approved):

PR #407, opened 2017-06-16
fix(notifications): handle slack errors silently

jrbotros:
Right now, Slack exceptions prevent submission of tasks, but we shouldn't be breaking on notifications in most cases.

NOTE: We might not want to ignore slack errors all the time? For example, if we're trying to add someone to a Slack group from the project management view we probably want to surface that error.

### DIFF
--- a/orchestra/communication/slack.py
+++ b/orchestra/communication/slack.py
@@ -1,18 +1,32 @@
 import random
 import string
 
+import slacker
 from django.conf import settings
 from django.utils.text import slugify
-from slacker import Slacker
 
 from orchestra.utils.decorators import run_if
 from orchestra.communication.errors import SlackFormatError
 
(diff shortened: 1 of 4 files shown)


orchestra/communication/slack.py
```diff
@@ -1,28 +1,40 @@
...
 
+import logging
+logger = logging.getLogger(__name__)
+
 # Types of responses we can send to slack
 VALID_RESPONSE_TYPES = {'ephemeral', 'in_channel'}
+_request = slacker.BaseAPI._request
+
+
+def _silent_request(*args, **kwargs):
```
marcua: docstring explaining this and a TODO saying we might want to surface certain errors

marcua (approved):

PR #413, opened 2017-07-14
Remove ng-annotate-loader, upgrade babel-plugin-angularjs-annotate

bgergen:


marcua (approved):
Approved assuming webpack doesn't create any files you need to use in addition.

PR #418, opened 2017-10-03
fixed iframe-wrapper style height error

devChuk:
The `iframe-wrapper` parent div of the iframe does not appropriately update its height when a height is passed into `website-iframe`. This leads to a UI bug where the iframe overflows out of its box in the UI.

This is a fix for that.

### DIFF
--- a/orchestra/static/orchestra/common/components/website-iframe/website-iframe.directive.es6.js
+++ b/orchestra/static/orchestra/common/components/website-iframe/website-iframe.directive.es6.js
@@ -32,6 +32,14 @@ export default function websiteIframe ($compile, $sce, $timeout) {
           iframe.setAttribute(attr, iframeAttrs[attr])
         }
 
+        var iframeWrappers = document.getElementsByClassName('iframe-wrapper')
+        for (var i = 0; i < iframeWrappers.length; i++) {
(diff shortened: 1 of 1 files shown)


orchestra/static/orchestra/common/components/website-iframe/website-iframe.directive.es6.js
```diff
@@ -32,6 +32,14 @@ export default function websiteIframe ($compile, $sce, $timeout) {
           iframe.setAttribute(attr, iframeAttrs[attr])
         }
 
+        var iframeWrappers = document.getElementsByClassName('iframe-wrapper')
+        for (var i = 0; i < iframeWrappers.length; i++) {
+          if (iframeWrappers[i].children[0].id === 'iframe-wrapper-' + $scope.id) {
```
marcua: is there any way to push this sort of logic into a selector so we don't need the `if`?

if not, then approve!:)

marcua (approved):
Approved, unless you can do something with the comment I left

thisisdhaas:
Also approved, assuming you've sanity-checked that this doesn't break our existing workflows. I'm probably satisfied if you check a designer and a CSM step for the launch_boost, website_enhancement, and recommendation_delivery workflows.

PR #421, opened 2017-10-11
Use isort to sort imports correctly

joshblum:
@marcua lmk what you think, i reccomend adding an [isort plugin](https://github.com/timothycrosley/isort/wiki/isort-Plugins) to your editor to automatically apply these rules during development. Would be good to automate these checks internally as well :)


marcua (approved):
Amazing thank you!

I just merged a monstrosity of a PR, so there are two small conflicts. Want to rerun and then merge?

PR #428, opened 2017-10-30
Separate beanstalk_dispatch to OSS

joshblum:
https://github.com/joshblum/beanstalk-dispatch ! @marcua you have write access and a shoutout in the readme :) We can also move this to the b12 org but just wanted to get it out of orchestra and on its own.


marcua:
@joshblum this all looks great! two thoughts:

- How do Orchestra users upgrade their codebase to support this? Presumably they don't need to update their urls.py because the dispatcher URLs look the same. Presumably they also don't have to update any `schedule_function` calls because the package didn't change.
- We'd love to have it in the b12io repo unless you like the glory!:) No strong feelings here so just let us know.

joshblum:
- I think that in order to upgrade the only change would be for  existing users to include `beanstalk-dispatch` in their `requirements.txt`. 

- I'm going to leave it under my name for now, but happy to move it over if you ever change your mind. Added you as an owner on pypi and you should have owner access on the repo.

marcua (approved):
Thank you for doing this! What a nice service to the community :).

Issue #434, opened 2018-01-02
SanityBot


marcua:
## Problem
When experts are working on tens of projects at once, it's easy to miss a step or lose situational awareness on a particular project.

## Solution
A SanityBot that periodically looks at the state of a project and reminds the project team about various things that seem off.

## Use cases
These use cases are heavily biased by the design delivery process at B12. While projects are more complex, we'll consider a project with a customer success manager (CSM) and a designer.
  * A CSM has been on a project for some time but has neither completed setting up the project or asked the customer for clarifying information.
  * The project is set up, but a designer has not been staffed to the project for some time.
  * A designer is staffed to a project, but we haven't sent v1 of a design to a customer in some time.
  * We've sent the v1 of a design to a customer, but haven't received confirmation/started iterating with the customer in some time.
  * We've been iterating with a customer for some time with no result.
  * We've done more than X hours of work on a project.
  * A todo hasn't been checked off in some time.

## Other features that are off the table for now
  * Projects/tasks should have "todo scaffolds," which are collections of todos that get created at various points in a project's lifetime.
  * Todos should have due dates, and should be snoozable for some amount of time before they pop up (reminders). If Todos could be set as reminders in some time, you could imagine not warning project teams about tasks while there is a snoozed Todo.
  * It would be nice to have a view of all of the Todos assigned to you across all projects, rather than having to click into individual tasks.

## How to surface sanity checks
  * Send a message to the project slack channel. Quick user research indicates that this is CSMs' preferred modality for surfacing checks.
  * Send a message to the internal (e.g., entire company) slack group/channel. This is for escalating sanity checks in cases where the project is very off track.
  * Create TODOs for team members (CSMs) when a sanity check occurs
  * Indicate tasks that have new sanity checks (Todos?) in the dashboard.

## Implementation notes
  * Create a `SanityCheck` model with the following fields
    * `created_at`: datetime
    * `is_deleted`: datetime
    * `handled_at`: datetime
    * `project`: foreign key
    * `check_slug`: a slug for the type of check that's fired. No uniqueness constraints on this---you might create multiple of the same sanity check every X hours/days.
  * In `version.json`, add an optional `sanity_checks` key with the following schema
    ```
    ...
    "sanity_checks": {
    "sanity_check_function": {
                  "path": "path.to.sanity.check.function"
    },
    "check_configurations": {
       "check_slug1": {
[...]

PR #435, opened 2018-01-04
SanityBot

marcua:
Addresses https://github.com/b12io/orchestra/issues/434


orchestra/migrations/0074_sanitycheck.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
marcua: won't merge this---will wait for @paopow 's PR, rebase, and rebuild migrations.

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+from orchestra.core.errors import SanityBotError
+from orchestra.models import Project
+from orchestra.models import SanityCheck
+from orchestra.models import WorkflowVersion
+from orchestra.utils.notifications import message_experts_slack_group
+from orchestra.utils.project_properties import incomplete_projects
+
+
+def _workflow_versions_with_sanity_checks():
+    # TODO(marcua): make more specific.
```
marcua: This doesn't incur serious performance issues in practice, so I'm tempted to leave as-is :)

Makefile
```diff
@@ -9,7 +9,9 @@ clean:
 	find . -name '*~' -delete
 
 lint:
-	flake8 . && isort --check-only --recursive .
+	# TODO(marcua): reenable isort when it works the same in dev and
+	# on CircleCI
+	flake8 . # && isort --check-only --recursive .
```
paopow: Yay!

orchestra/migrations/0074_sanitycheck.py
paopow: It's fine really. :-D

orchestra/models/core/models.py
```diff
@@ -64,13 +65,16 @@ class WorkflowVersion(WorkflowVersionMixin, models.Model):
             A longer description of the workflow version.
         workflow (orchestra.models.Workflow):
             The workflow that this is a version of.
+        sanity_checks (str):
+            A JSON blob used to define the sanity checks we run.
```
paopow, line 69: Where do I find an example of the format of this blob?

orchestra/tests/helpers/workflow.py
```diff
@@ -552,6 +557,56 @@ def machine_task_function():
...
+                    'check_configurations': {
+                        "friendly_check": {
+                            "handlers": [{
+                                "type": "slack_project_channel",
+                                "message": "Friendly message",
+                                "steps": ["step1"]
+                            }],
+                            "repetition_seconds": 86400
+                        },
+                    }
```
paopow: Can we use this as an example of the sanity_checks blob? What would it look like with multiple checks? What happens to "deleted" checks?

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+    message_experts_slack_group(
+        project.slack_group_id, message)
+
+
+def _filter_checks(project, checks, check_configurations):
+    latest_check_creation = (SanityCheck.objects
+                             .filter(project=project)
+                             .values('check_slug')
+                             .annotate(max_created_at=Max('created_at')))
+    latest_check_creation = {
```
paopow: Reusing the variable name is OK?

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+        check['check_slug']: check['max_created_at']
+        for check in latest_check_creation}
+    for check in checks:
+        max_created_at = latest_check_creation.get(check.check_slug)
```
paopow: What is the different between check['check_slug'] and check.check_slug?

Makefile
paopow: Yay

orchestra/migrations/0074_sanitycheck.py
paopow: It's fine really. :D

orchestra/models/core/models.py
```diff
@@ -64,13 +65,16 @@ class WorkflowVersion(WorkflowVersionMixin, models.Model):
...
     """
     created_at = models.DateTimeField(default=timezone.now)
     slug = models.CharField(max_length=200)
     name = models.CharField(max_length=200)
     description = models.TextField()
     workflow = models.ForeignKey(
         Workflow, related_name='versions', on_delete=models.CASCADE)
+    sanity_checks = JSONField(default={})
```
paopow, line 80: Why storing a lot of information here instead of the SanityCheck itself?

orchestra/bots/sanitybot.py
paopow: Same variable name OK?

orchestra/bots/sanitybot.py
paopow: What is the difference between check['check_slug'] and check.check_slug?

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+    for project in incomplete_projects(
+            Project.objects.filter(
+                workflow_version__in=workflow_versions)):
+        sanity_checks = project.workflow_version.sanity_checks
+        sanity_check_path = (sanity_checks
+                             .get('sanity_check_function', {})
+                             .get('path'))
+        check_configurations = sanity_checks.get('check_configurations')
+        if sanity_check_path and check_configurations:
+            sanity_check_function = locate(sanity_check_path)
```
paopow, line 83: What if locate fail?

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+            sanity_checks = sanity_check_function(project)
```
paopow, line 84: What if the function return an invalid value?

orchestra/bots/sanitybot.py
```diff
@@ -0,0 +1,92 @@
...
+            raise SanityBotError(
+                'No handlers for {}'.format(sanity_check.check_slug))
+        for handler in handlers:
+            _handle(project, sanity_check, handler)
+        sanity_check.handled_at = timezone.now()
+        sanity_check.project = project
+        sanity_check.save()
+
+
+def create_and_handle_sanity_checks():
```
paopow, line 72: In a deployed instance, when is this function called? How do we automatically check and handle the SanityCheck?

orchestra/bots/sanitybot.py
marcua: it's nasty! fixed!:)

orchestra/bots/sanitybot.py
marcua: `check` is of type `SanityCheck` (we both miss typed languages :)), which has a property called `check_slug`

`latest_check_creation` is a dictionary, which has keys defined by the set of distinct `check_slug`

orchestra/bots/sanitybot.py
marcua: it will throw an exception, which will stop the world. this has been how we've handled all `*_function` paths.

longer-term, we should consider 'statically analyzing' workflow configurations on load.

orchestra/bots/sanitybot.py
marcua: in general it gets "sanity checked" in  `_handle_sanity_checks(`. Can you think of ways it'd break?

orchestra/models/core/models.py
marcua: reasonable question, updated docstring

orchestra/bots/sanitybot.py
marcua: Added docs!

paopow (approved):

PR #437, opened 2018-01-24
Team info directive

paopow:
Add a team info card directive that show a table of team members, their respective roles and the time they have spend on the project.

### DIFF
--- a/orchestra/project_api/serializers.py
+++ b/orchestra/project_api/serializers.py
@@ -1,3 +1,4 @@
+import datetime
 from rest_framework import serializers
 
 from orchestra.models import Iteration
@@ -104,7 +105,7 @@ class TaskSerializer(serializers.ModelSerializer):
 
 
 class TaskAssignmentSerializer(serializers.ModelSerializer):
-
+    recorded_work_time = serializers.SerializerMethodField()
     class Meta:
         model = TaskAssignment
         fields = (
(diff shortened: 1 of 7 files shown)


orchestra/project_api/serializers.py
```diff
@@ -154,6 +157,17 @@ def get_in_progress_task_data(self, obj):
...
+    def get_recorded_work_time(self, obj):
+        query = TimeEntry.objects.filter(worker=obj.worker, assignment=obj)
+        if not query:
+            return None
+
+        total_time = datetime.timedelta()
+        for t in query:
+            total_time += t.time_worked
+
+        return str(total_time)
```
marcua: if it's not a huge amount of work, return `total_seconds` or something so we can decide what to do with it on the frontend?

if it's a huge amount of work, don't do it :)

marcua: given you're using moment.js on the frontend, this seems like not a huge amount of work :)

orchestra/static/orchestra/team-info/team-info-card.directive.es6.js
```diff
@@ -0,0 +1,36 @@
...
+        .then(response => {
+          const {steps, tasks} = response.data
+          const humanSteps = new Set(steps.filter(step => step.is_human).map(step => step.slug))
+          teamInfoCard.assignments = []
+          for (let stepSlug of humanSteps.values()) {
+            const task = tasks[stepSlug]
+            if (task) {
+              teamInfoCard.assignments = teamInfoCard.assignments.concat(task.assignments.map(a => {
+                return {
+                  role: stepSlug,
```
marcua: use the pretty version like we have in todos

marcua (changes requested):
back atcha

marcua (approved):

PR #442, opened 2018-02-27
Upgrade django to 1.11 LTS

lydiagu:


orchestra/utils/task_lifecycle.py
```diff
@@ -745,7 +745,7 @@ def get_new_task_assignment(worker, task_status):
              .order_by('project__start_datetime'))
 
     certification_error = False
-    for task in tasks.iterator():
+    for task in tasks:
```
lydiagu, line 748: Django 1.11 uses server-side cursors for iterator(), which caused errors in the unit tests https://docs.djangoproject.com/en/2.0/releases/1.11/#database-backends. I figured removing the `.iterator()` would keep the same behavior as before 1.11

paopow (approved):

marcua (approved):
a

PR #449, opened 2018-06-12
Checklist models

paopow:
Update the Todo model and add the ChecklistTemplate model.


orchestra/models/core/models.py
```diff
@@ -607,6 +646,12 @@ class Meta:
     completed = models.BooleanField(default=False)
     start_by_datetime = models.DateTimeField(null=True, blank=True)
     due_datetime = models.DateTimeField(null=True, blank=True)
+    skipped = models.BooleanField(default=False)
+    parent_todo = models.ForeignKey(
+        'self', null=True, related_name='parent')
+    template = models.ForeignKey(
```
marcua, line 652: whenever null=True, also do blank=True so you can save it in admin :)

orchestra/models/core/models.py
```diff
@@ -607,6 +646,12 @@ class Meta:
     completed = models.BooleanField(default=False)
     start_by_datetime = models.DateTimeField(null=True, blank=True)
     due_datetime = models.DateTimeField(null=True, blank=True)
+    skipped = models.BooleanField(default=False)
```
marcua, line 649: skipped_datetime so we can reconstruct when these happened :)

orchestra/models/core/models.py
```diff
@@ -607,6 +646,12 @@ class Meta:
     completed = models.BooleanField(default=False)
```
marcua, line 646: we could consider completed_datetime, but that could be a separate migration

orchestra/models/core/models.py
```diff
@@ -607,6 +646,12 @@ class Meta:
...
     start_by_datetime = models.DateTimeField(null=True, blank=True)
     due_datetime = models.DateTimeField(null=True, blank=True)
+    skipped = models.BooleanField(default=False)
+    parent_todo = models.ForeignKey(
+        'self', null=True, related_name='parent')
+    template = models.ForeignKey(
+        ChecklistTemplate, null=True, related_name='template')
+    log = JSONField(default={'actions': []})
```
marcua, line 654: activity_log ?

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
     end_date = models.DateField(null=True, blank=True)
 
 
+class ChecklistTemplate (ChecklistTemplateMixin, BaseModel):
```
marcua, line 580: Given the objects are called Todos, can we call this TodoListTemplate and TotoListTemplateMixin?

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+    """
+    A checklist template
+
+    Attributes:
+        slug (str):
+            The string that use to refer the checklist
```
marcua, line 586: A unique key for the todo list

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+        name (str):
+            The name of the checklist
```
marcua, line 588: A human-readable name for the todo list

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+        description (str):
+            A text description of the checklist
```
marcua, line 590: text -> longer

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+        creator (orchestra.models.Worker)
+            The worker who creates the checklist. This field is null
```
marcua, line 592: creates -> created

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+    """
+    class Meta:
+        app_label = 'orchestra'
+
+    slug = models.CharField(max_length=200, unique=True)
+    name = models.CharField(max_length=200)
+    description = models.TextField()
+    creator = models.ForeignKey(
+        Worker, null=True, related_name='creator', on_delete=models.CASCADE)
+    todos = JSONField(default={'list': []})
```
marcua, line 605: document the format

orchestra/models/core/models.py
```diff
@@ -576,6 +577,34 @@ class PayRate(PayRateMixin, models.Model):
...
+            A JSON blob that describe the todos in the checklist
+    """
+    class Meta:
+        app_label = 'orchestra'
+
+    slug = models.CharField(max_length=200, unique=True)
+    name = models.CharField(max_length=200)
+    description = models.TextField()
+    creator = models.ForeignKey(
+        Worker, null=True, related_name='creator', on_delete=models.CASCADE)
```
marcua, line 604: also blank=True

marcua, line 604: should the delete SET_NULL so we don't lose the template?

orchestra/models/core/models.py
adbharadwaj: @marcua A few doubts. :)
- Arent we planning to use activity log to reconstruct what happened?
- What happens when a user skips and then adds back? I suppose we set the field to null when someone unskips a todo.

orchestra/models/core/models.py
adbharadwaj: I like `activity_log`.

orchestra/models/core/models.py
adbharadwaj: @marcua Shouldn't manage different versions of a template (add a version column?). I suppose we are going to update templates over time.

marcua: I'm pro-versioning, though I think it's too early for the needs of the product right now. I still think we should SET_NULL to be safe for the time being.

orchestra/models/core/models.py
marcua: In general, the best solution for activity logging is to not write it out to the DB, but instead to event log into tools like Snowplow or Google Analytics. So I'm a little skeptical of the activity_log in the long-term, and am a big fan of skipped_datetime as an imperfect signal.

Setting to null if unskipped seems great!

adbharadwaj: Cool :) I am a big fan of Google analytics event tags too :)

PR #453, opened 2018-06-20
Backend support for todolists and todolist templates

adbharadwaj:
The PR does the following things:
- Added a function to transform the todo list results into nested json for tree view (on frontend). 
- Added create, retrieve and update functions for todolist templates.
- Added a POST api `\orchestra\todos\add_todos_from_todolist_template` to add items from template to the tasks todolist.
    - The api request requires two parameters - `task` and `todolist_template`
    - The api response returns the updated list of todos.
- Register an admin panel for `TodoListTemplate`
- Note: Migration script was updated because I was trying to test the use of double qoutes in todos JsonField in `TodoListTemplate`.


orchestra/migrations/0076_auto_20180621_1524.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
```
marcua: We never change migrations across PRs. If the old one was committed to master, you should create a new one.

orchestra/todos/views.py
```diff
@@ -91,3 +74,17 @@ def perform_update(self, serializer):
...
+
+
+class TodoListTemplateDetail(generics.RetrieveUpdateAPIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    serializer_class = TodoListTemplateSerializer
+    queryset = TodoListTemplate.objects.all()
+
+
+class TodoListTemplateList(generics.ListCreateAPIView):
```
marcua, line 96: we probably just want list, but not create?

orchestra/todos/views.py
```diff
@@ -91,3 +74,17 @@ def perform_update(self, serializer):
             'complete' if todo.completed else 'incomplete')
         message_experts_slack_group(
             todo.task.project.slack_group_id, message)
+
+
+class TodoListTemplateDetail(generics.RetrieveUpdateAPIView):
```
marcua, line 89: do we need the detail view, or do we get the details in the list already?

orchestra/todos/decorators.py
```diff
@@ -0,0 +1,33 @@
+import logging
+
+from django.views.decorators.csrf import csrf_exempt
+from jsonview.decorators import json_view
+from rest_framework.decorators import api_view
+# from rest_framework.decorators import authentication_classes
```
marcua: :)

orchestra/todos/decorators.py
```diff
@@ -0,0 +1,33 @@
+import logging
```
marcua: 1) This feels more generic than just `todos`
2) Do we need this? We've already got a pattern for decorating API requests.

orchestra/todos/views.py
```diff
@@ -1,47 +1,30 @@
...
-            return worker.assignments.filter(task__project=project_id).exists()
-        elif request.method == 'POST':
-            # Create calls have a task ID
-            task_id = request.data.get('task')
-            project = Task.objects.get(id=task_id).project
-            return worker.assignments.filter(task__project=project).exists()
-        return False
+@api_endpoint(['POST'])
+def add_todos_from_todolist_template(request):
+    todolist_template_id = request.data.get('todolist_template')
```
marcua: if we're using template ids, we don't need a slug :).

if we use the slug, we should make sure it's db_indexed

orchestra/migrations/0076_auto_20180621_1524.py
adbharadwaj: @marcua Do you mean I should create a migration on top of the old one or delete the old migration and create a new one.

Btw for some reason, Github is showing the diff between two different files here. Instead of showing two different changes (1 file created and 1 file deleted).

orchestra/todos/views.py
adbharadwaj: @marcua Create will be required to allow designers to save todolist templates from the user interface.

orchestra/todos/views.py
adbharadwaj: @marcua Retrieve may not be necessary at this point. But we will need to have `Update` to allow designers to save their custom todolist. 

Retrieve may come in handle later on when designers want to check out other designers todolist in a different interface.

orchestra/todos/decorators.py
adbharadwaj: @marcua Maybe I am using a really old api as my reference - https://github.com/b12io/orchestra/tree/master/orchestra/project_api 

Can you point me to the pattern for decorating API requests?

orchestra/todos/api.py
```diff
@@ -0,0 +1,32 @@
...
+from orchestra.models import Task
+
+logger = logging.getLogger(__name__)
+
+
+def add_todolist_template(todolist_template_id, task_id):
+    todolist_template = TodoListTemplate.objects.get(id=todolist_template_id)
+    task = Task.objects.get(id=task_id)
+    template_todos = todolist_template.todos['items']
+    parent_todo = Todo(
```
paopow: `root_todo` to differentiate from normal parents?

orchestra/tests/test_todos.py
```diff
@@ -161,11 +167,174 @@ def test_create_todo_with_due_datetime(self):
...
+            _todo_data(self.task, 'todo parent', False,
+                       template=todolist_template.id,
+                       parent_todo=todos[2]['id']),
+            _todo_data(self.task, self.todolist_template_name,
+                       False, template=todolist_template.id),
+        ]
+        for todo, expected_todo in zip(todos, expected_todos):
+            self._verify_todo_content(todo, expected_todo)
+
+    def test_add_todos_from_todolist_template_forbidden(self):
```
paopow: So this is forbidden because there is no task_id?

orchestra/migrations/0076_auto_20180621_1524.py
marcua: Once a migration is committed to master, you should assume that someone has run it, which means it shouldn't change. If you want to subsequently modify a models.py (that's a great thing!:)), then create a new migration (e.g., 0077 in this case) so that anyone who has already applied 0076 doesn't end up in a broken state.

orchestra/todos/views.py
marcua: I suspect we're overengineering a little since experts won't be able to update or retrieve other experts' lists in v1. I don't feel too strongly about this if you want to leave it in and if you've tested the endpoint :).

orchestra/todos/decorators.py
marcua: Given that https://github.com/b12io/orchestra/tree/master/orchestra/project_api/decorators.py already exists, you shouldn't duplicate it. Either import from it, or move it to a more central place and import in both locations.

If there are any other copy/pasted modules in this PR, do the same for them.

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -101,6 +101,16 @@ export default function todoList (orchestraApi) {
         $scope.$apply()
       }
 
+      todoList.transformToTree = (todos) => {
+        var nodes = {}
+        return todos.filter(function (obj) {
+          nodes[obj.id] = defaults(obj, nodes[obj.id], { items: [] })
+          obj.parent_todo && (nodes[obj.parent_todo] = (nodes[obj.parent_todo] || { items: [] }))['items'].push(obj)
+
+          return !obj.parent_todo
```
paopow, line 110: Sanity check: return [] when `todos` is empty is the desired behavior?

orchestra/todos/api.py
```diff
@@ -0,0 +1,32 @@
...
+from orchestra.models import Todo
+from orchestra.models import Task
+
+logger = logging.getLogger(__name__)
+
+
+def add_todolist_template(todolist_template_id, task_id):
+    todolist_template = TodoListTemplate.objects.get(id=todolist_template_id)
+    task = Task.objects.get(id=task_id)
+    template_todos = todolist_template.todos['items']
```
paopow: How does this handle invalid `todolist_template_id` and `task_id`?

orchestra/tests/test_todos.py
adbharadwaj: @paopow Yes. This means you cannot add todo without an associated task.

orchestra/todos/api.py
adbharadwaj: :)

orchestra/todos/views.py
adbharadwaj: I want to leave it in. Especially since I have written the tests. :)

orchestra/static/orchestra/todos/todo-list.directive.es6.js
adbharadwaj: if todos is [], transformToTree returns []

paopow (approved):

orchestra/project_api/views.py
```diff
@@ -12,14 +13,19 @@
...
+from orchestra.project_api.auth import IsSignedUser
 
 logger = logging.getLogger(__name__)
 
 
-@api_endpoint(['POST'])
+@api_endpoint(methods=['POST'],
+              permissions=(IsSignedUser,),
+              logger=logger)
+@authentication_classes((OrchestraProjectAPIAuthentication,))
```
marcua: why not push optional auth classes into the endpoint decorator?

orchestra/tests/helpers/fixtures.py
```diff
@@ -169,6 +169,20 @@ class Meta:
...
 
+class TodoListTemplateFactory(factory.django.DjangoModelFactory):
+    slug = factory.Sequence(
+        lambda n: 'todolist_template_{}'.format(n))
+    name = factory.Sequence(
+        lambda n: 'TodoList template {}'.format(n))
+    description = factory.Sequence(
+        lambda n: 'Description {}'.format(n))
+    creator = None
+    todos = "{'items': []}"
```
marcua: string not dict?

orchestra/utils/decorators.py
```diff
@@ -1,4 +1,54 @@
...
+    def func_wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            logger.exception('An API exception occurred')
+            raise e
+    return func_wrapper
+
+
+def api_endpoint(methods, permissions, logger):
```
marcua: do you want to create two decorators, one which includes this and one which includes login_required and one which includes auth classes? that way you can use one decorator instead of two per view?

marcua (approved):

PR #454, opened 2018-06-22
Frontend support for adding template todo list items

adbharadwaj:
The PR implements the select box for templates.
screen shot 2018-06-26 at 4 59 42 pm

The PR also implements A/B testing feature for template todo list. In the implementation, any user who is part of the `todolist_feature` group will be able to see the template related feature. Here is a screenshot of the interface for the users who are not part of the todo list template feature group.
screen shot 2018-06-26 at 4 55 25 pm


orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -76,6 +87,16 @@ export default function todoList (orchestraApi) {
         })
       }
 
+      todoList.addTodoListTemplate = () => {
```
marcua: is this redundant with `addTodosFromTodoListTemplate` Do we only need one?

orchestra/static/orchestra/todos/todolist-template.service.es6.js
```diff
@@ -0,0 +1,18 @@
+export default function todoListTemplateApi ($http) {
+  const apiBase = '/orchestra/todos/'
+
+  const list = () => {
+    return `${apiBase}todolist_templates/`
+  }
+
+  const addTodoListTemplate = () => {
```
marcua: high-level throughout: should we be "creating/updating todo list from template" or "adding todo list template?" My vote is the first one, meaning in this example it's `createTodoListFromTemplate` or `updateTodoListFromTemplate`

marcua (changes requested):
small things

orchestra/static/dist/main.css
```diff
@@ -2345,6 +2345,8 @@ a.logo {
   .list-by-status .angular-ui-tree-node {
```
marcua, line 2349: unrelated to this PR: by default, does the `newTodoTaskId` get set to the current task's value, or will one expert's dropdown be set to another expert's `newTodoTaskId` by default?

orchestra/static/orchestra/todos/todolist-template.service.es6.js
adbharadwaj: I like `updateTodoListFromTemplate`! It clear explains that we are updating TodoList. But in such a case I would prefer to change the API URL to `update_todos_from_todolist_template`.  What do you think?

I dont like `createTodoListFromTemplate` because we are actually updating the todos from the template.

orchestra/static/orchestra/todos/todo-list.directive.es6.js
adbharadwaj: Well, I think I over-engineered for a scenario when `addTodosFromTodoListTemplate` might be reused. I cannot think of any at this point. I will just make it one function for now.

orchestra/static/dist/main.css
adbharadwaj: One expert's dropdown can be set to another expert's newTodoTaskId.

orchestra/static/dist/main.css
```diff
@@ -2309,16 +2309,20 @@ a.logo {
     background-color: white;
```
paopow, line 2309: Add a screenshot that shows the drop down list?

paopow (approved):

orchestra/todos/views.py
```diff
@@ -98,3 +98,11 @@ class TodoListTemplateList(generics.ListCreateAPIView):
 
     serializer_class = TodoListTemplateSerializer
     queryset = TodoListTemplate.objects.all()
+
+    def get_queryset(self):
+        if self.request.user.groups.filter(name='todolist_feature').exists():
```
marcua: `todolist_template_feature`

marcua:
```
# Only enable todolist template feature functionality for users in the `todolist_template_feature` group to facilitate A/B testing.
```

orchestra/static/dist/main.css
```diff
@@ -2309,16 +2309,20 @@ a.logo {
...
     text-align: center; }
```
marcua, line 2311: agree on `update_todos_from_todolist_template`

marcua (approved):
a, but `update_todos_from_todolist_template` and change feature group name. no need for another review if you do those things!:)

PR #456, opened 2018-06-28
Dynamic todo template

paopow:
Add an option to specify todos to remove or skip from a todo list template based on computed conditional properties.


orchestra/models/core/models.py
```diff
@@ -592,6 +592,9 @@ class TodoListTemplate (TodoListTemplateMixin, BaseModel):
             if it is generated by the system.
         todos (str)
             A JSON blob that describe the todos in the todo list template.
+        conditional_property_function (str)
+            A JSON blob containing the path to and name of a python method
+            that will return the preconditions to prune the created todos
```
marcua: nitpick: period at end of sentence

orchestra/todos/api.py
```diff
@@ -17,17 +21,50 @@ def add_todolist_template(todolist_template_slug, task_id):
...
-    todo = Todo(
-        task=task,
-        description=template_todo['description'],
-        template=todolist_template,
-        parent_todo=parent_todo
-    )
-    todo.save()
-    for template_todo_item in template_todo.get('items', []):
-        _add_template_todo(template_todo_item, todolist_template, todo, task)
+def _to_exclude(props, conditions=[]):
```
marcua: https://stackoverflow.com/questions/5712904/empty-dictionary-as-default-value-for-keyword-argument-in-python-function-dicti :)

you can do this:
```
def _to_exclude(props, conditions=None):
    for condition in conditions or []:
```

orchestra/todos/api.py
```diff
@@ -17,17 +21,50 @@ def add_todolist_template(todolist_template_slug, task_id):
...
+
+    cond_props = {}
+    path = todolist_template.conditional_property_function.get(
+        'path', None)
+    if path:
+        try:
+            get_cond_props = locate(path)
+            cond_props = get_cond_props(task.project)
+        except Exception:
+            logger.exception('Invalid condition function path.')
```
marcua: condition->contional

orchestra/todos/api.py
marcua: alternatively, you already pass in `[]` below, so just make `conditions` required ;)

orchestra/models/core/models.py
```diff
@@ -592,6 +592,9 @@ class TodoListTemplate (TodoListTemplateMixin, BaseModel):
             if it is generated by the system.
         todos (str)
             A JSON blob that describe the todos in the todo list template.
+        conditional_property_function (str)
```
marcua, line 595: explain the structure---is it a list of conditions that get ORed together, with properties in each dictionary getting ANDed, or the other way?

orchestra/todos/api.py
marcua: docstring to rehash this question above:

explain the structure---is it a list of conditions that get ORed together, with properties in each dictionary getting ANDed, or the other way?

orchestra/todos/api.py
```diff
@@ -17,17 +21,50 @@ def add_todolist_template(todolist_template_slug, task_id):
...
-    todo.save()
-    for template_todo_item in template_todo.get('items', []):
-        _add_template_todo(template_todo_item, todolist_template, todo, task)
+def _to_exclude(props, conditions=[]):
+    for condition in conditions:
+        for prop, predicate in condition.items():
+            current_value = props.get(prop)
+            compared_to_value = predicate['value']
+            if predicate['operator'] == '=':
+                return current_value == compared_to_value
```
marcua: since these return immediately, they only ever check the first property of the first condition. you probably want to keep a boolean around that you keep flipping depending on properties/conditions

marcua (changes requested):
A few things

orchestra/todos/api.py
adbharadwaj: Another way to implement string operators -> https://stackoverflow.com/questions/44554614/eval-string-cmp-operator

This might be useful if we intend to extend the available operators in future.

orchestra/todos/api.py
```diff
@@ -17,17 +21,50 @@ def add_todolist_template(todolist_template_slug, task_id):
...
+    if not remove:
+        to_skip = _to_exclude(
+                conditional_props, template_todo.get('skip_if', []))
+        skipped_datetime = timezone.now() if to_skip else None
+        todo = Todo(
+            task=task,
+            description=template_todo['description'],
+            template=todolist_template,
+            parent_todo=parent_todo,
+            skipped_datetime=skipped_datetime
```
adbharadwaj, line 89: To propogate skipping from parents to descendents.
```python
skipped_datetime = timezone.now() if parent_todo.skipped_datetime or to_skip else None
```
We may even want to skip evaluation of `to_skip` if parent_todo.skipped_datetime is not Null. That may save considerable computation time.

adbharadwaj (commented):
A small suggestion regarding the string operators and a comment on skip propagation.

orchestra/todos/api.py
```diff
@@ -30,32 +40,45 @@ def add_todolist_template(todolist_template_slug, task_id):
...
     for template_todo in template_todos:
         _add_template_todo(
             template_todo, todolist_template, root_todo, task, cond_props)
 
 
-def _to_exclude(props, conditions=[]):
+def _to_exclude(props, conditions):
+    """
+    The conditions is it a list of conditions that get ORed together,
+    with properties in each dictionary getting ANDed.
```
marcua: with predicates in each dictionary

orchestra/todos/api.py
```diff
@@ -30,32 +40,45 @@ def add_todolist_template(todolist_template_slug, task_id):
...
+    """
+    any_condition_true = False
+
     for condition in conditions:
+        all_props_true = True
```
marcua: are we fine with a zero-item list of properties evaluating to true? could do `all_props_true = True and len(all_props_true) > 0`

orchestra/todos/api.py
```diff
@@ -25,36 +26,37 @@ def add_todolist_template(todolist_template_slug, task_id):
...
     """
     The conditions is it a list of conditions that get ORed together,
-    with properties in each dictionary getting ANDed.
+    with predicates in each dictionary getting ANDed.
     """
     any_condition_true = False
 
     for condition in conditions:
-        all_props_true = True
+        all_props_true = len(condition) > 0
```
marcua: len(conditions) :P

orchestra/todos/api.py
```diff
@@ -25,36 +26,37 @@ def add_todolist_template(todolist_template_slug, task_id):
 
     task = Task.objects.get(id=task_id)
     template_todos = todolist_template.todos.get('items', [])
-    root_todo = Todo(
-        task=task,
-        description=todolist_template.name,
-        template=todolist_template
-    )
-    root_todo.save()
+    with transaction.atomic():
```
marcua: you could wrap the entire function in `@transaction.atomic` and then you don't have to indent everything :)

orchestra/json_schemas/todos.py
```diff
@@ -1,10 +1,54 @@
...
+    A predicate schema
+
+    Attributes:
+        operator (str):
+            Specify the compare operator of predicate.
+            Supported predicates are >, <, >=, <=, !=, and ==.
+        value (number, bool, sring, or None):
+            The value to compare with of the predicate.
+    """
+    operator = jsl.StringField(required=True, pattern='[!=><]=|[><]')
```
marcua, line 15: `[!=><]=|[><]` looks like a serious Kirby battle

orchestra/todos/api.py
paopow: This is for detecting {} (a condition). if len(conditions) == 0, this line wouldn't get executed.

marcua (approved):
[!=><]=|[><]

PR #460, opened 2018-07-13
Reduces slack messages on project channel when experts use todolists.

adbharadwaj:
This PR makes two changes:

1. Send messages to slack only when:
     - [x] Todo item has no parent (todos created by the worker or root level template todos)
     - [x] Todo has a parent with no parents (todo items at depth 1)
2. Show a different message when a user skips/un-skips a todo in comparison to marking it as complete/incomplete.

### DIFF
--- a/orchestra/todos/views.py
+++ b/orchestra/todos/views.py
@@ -134,8 +134,12 @@ class TodoDetail(generics.RetrieveUpdateDestroyAPIView):
             todo_change = 'not relevant' \
                 if todo.skipped_datetime else 'relevant'
         else:
+            # When activity_log is updated, `todo_change = None`
+            # to avoid triggering any slack messages
             todo_change = None
 
+        # To avoid Slack noise, only send updates for changed TODOs with
+        # depth 0 (no parent) or 1 (no grantparent).
         if todo_change and \
                 (not (todo.parent_todo and todo.parent_todo.parent_todo)):
             message = '{} has marked `{}` as `{}`.'.format(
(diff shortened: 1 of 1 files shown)


orchestra/todos/views.py
```diff
@@ -123,15 +123,27 @@ class TodoDetail(generics.RetrieveUpdateDestroyAPIView):
...
-        message_experts_slack_group(
-            todo.task.project.slack_group_id, message)
+
+        if old_todo.completed != todo.completed:
+            todo_change = 'complete' if todo.completed else 'incomplete'
+        elif old_todo.skipped_datetime != todo.skipped_datetime:
+            todo_change = 'not relevant' \
+                if todo.skipped_datetime else 'relevant'
+        else:
+            todo_change = None
```
marcua: Should this ever happen? If not, throw a logger.warning on it.

orchestra/todos/views.py
```diff
@@ -123,15 +123,27 @@ class TodoDetail(generics.RetrieveUpdateDestroyAPIView):
...
+
+        if todo_change and \
```
marcua: `# To avoid Slack noise, only send updates for changed TODOs with depth 0 (no parent) or 1 (no grantparent).`

marcua (changes requested):
One question, one comment

orchestra/todos/views.py
marcua: Ohh...It will happen when we add activity logging, eh?

adbharadwaj: Right! when we add activity logging.

marcua (approved):

PR #461, opened 2018-07-13
Add todos activity logging

adbharadwaj:
Here is an example of how the activity log will look like:
```json
{
   "actions":[
      {
         "action":"skip",
         "datetime":"2018-07-13 20:22:01"
      },
      {
         "action":"unskip",
         "datetime":"2018-07-13 20:22:08"
      },
      {
         "action":"expand",
         "datetime":"2018-07-13 20:22:51"
      },
      {
         "action":"collapse",
         "datetime":"2018-07-13 20:22:53"
      },
      {
         "action":"complete",
         "datetime":"2018-07-13 20:22:55"
      },
      {
         "action":"incomplete",
         "datetime":"2018-07-13 20:22:55"
      }
   ]
}
```


orchestra/todos/serializers.py
```diff
@@ -39,12 +43,16 @@ class Meta:
             'completed',
             'skipped_datetime',
             'start_by_datetime',
-            'due_datetime')
+            'due_datetime',
+            'activity_log')
         read_only_fields = ('id',)
 
 
 class TodoWithQASerializer(serializers.ModelSerializer):
```
marcua, line 51: Do we want to log activity on the TodoWithQA? Or alternatively, do we want to log activity on TodoWithoutQA?:)

orchestra/tests/test_todos.py
```diff
@@ -24,7 +24,8 @@
 
 def _todo_data(task, description, completed,
                skipped_datetime=None, start_by=None,
-               due=None, parent_todo=None, template=None, qa=None):
+               due=None, parent_todo=None, template=None,
+               activity_log=str({'actions': []}), qa=None):
```
marcua, line 28: Is `json.dumps` preferable? For example, I think that `'` isn't a valid JSON quote.

orchestra/static/orchestra/todos/todo-checklist.html
```diff
@@ -5,7 +5,7 @@
            ng-model="todo.completed"
            ng-change="updateTodo(todo)">
     <label ng-class="{'text-danger': isInDanger(todo)}">
-      <span ng-click="toggle(this)">
+      <span ng-click="toggleTodo(todo, this)">
```
marcua, line 8: two things:
1) My mind is cloudy, but why is it possible to toggle a checklist item from two places (`todo-checklist.html` and `todo-list.html`
2) Should this file be called `todo-list-item.html`?

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -98,9 +99,26 @@ export default function todoList (orchestraApi) {
...
+        activityLog['actions'].push({
+          'action': action,
+          'datetime': datetimeUtc.format('YYYY-MM-DD HH:mm:ss')
+        })
+        todo.activity_log = JSON.stringify(activityLog).replace(/'/g, '"')
+      }
+
+      todoList.onToggleTodo = (todo, collapsed) => {
+        todoList.addActionToTodoActivityLog(todo, collapsed ? 'collapse' : 'expand')
+        todoApi.update(todo)
```
marcua, line 115: given the messages happen asynchronously, it's possible that activity log [a,b,c] arrives before [a,b] and we lose c.

are we OK with that? if not, do we want to add logic on the backend to only log appends to the activity log? i'm personally OK losing a few activities---just wanted to raise the idea that we might :)

orchestra/static/orchestra/todos/todo-list.html
```diff
@@ -100,6 +101,7 @@ <h3>
               skip-todo="todoList.skipTodo"
               unskip-todo="todoList.unskipTodo"
               task-slugs="todoList.taskSlugs"
+              on-toggle-todo="todoList.onToggleTodo"
```
marcua, line 104: consistency: the rest of them  aren't preceded wtih "on-" :)

marcua (changes requested):
A few comments

orchestra/static/orchestra/todos/todo-list.html
adbharadwaj: The rest of them are actually taking actions like "skip" or "unskip" whereas `on-toggle-todo` is an event handler. It does not toggle the todo. It is handling the event when a todo is toggled.

orchestra/todos/serializers.py
adbharadwaj: `TodoWithoutQA` or `TodoWithQA` is just a way to make sure that QA data is only visible to the assigned worker. I would say both.

orchestra/static/orchestra/todos/todo-list.directive.es6.js
adbharadwaj: Technically yes that can happen. But on the interface ideally, it should happen since the user can only take one action at a time. For example, an item cannot be toggled and skipped at the same time. I would say the time lag to update activity log should be very low. 

I am OK with that.

orchestra/static/orchestra/todos/todo-checklist.html
adbharadwaj: 1. The functionality to toggle is implemented in `todo-checklist` whereas functionality of what happens when a todo is toggled is implemented in `todo-list`.
2. This file not only renders the entire todolist in a checklist format and hence the name `todo-checklist.html`

orchestra/tests/test_todos.py
adbharadwaj: Tried `json.dumps` but it causes the tests to fail because `json.dumps({'actions': []})` equates to `{"actions": []}` whereas ORM returns it as "{'actions': []}".

A possible solution is to create a SerializerMethodField which performs `json.dumps` on `activity_log`. What do you think?

orchestra/todos/serializers.py
marcua: I'm saying we shouldn't log activity not attributable the assigned worker. For example, in the current design would we end up logging a reviewer's expand/collapse activity?

orchestra/tests/test_todos.py
marcua: If you can find an example of this being the right way to do things™, I'd be down. If you can't I'd say it's not worth worrying about too much given it's in tests :)

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -98,9 +99,26 @@ export default function todoList (orchestraApi) {
...
+      }
+
       todoList.skipTodo = (todo) => {
         const datetimeUtc = moment.tz(moment(), moment.tz.guess()).utc()
         todo.skipped_datetime = datetimeUtc.format('YYYY-MM-DD HH:mm')
+        // Not passing todo.skipped_datetime because we want to log datetime at a higher resolution (includes seconds).
```
marcua: that's funny---I'd have argued we should have the same high-resolution data in both :)

adbharadwaj: Sanity check
- [x] Old skipped todos works fine.
- [x] Due date functionality works fine.

orchestra/tests/test_todos.py
adbharadwaj: Cant find an example. As you say, I wouldn't worry about it. :)

orchestra/todos/serializers.py
adbharadwaj: Adding step slug to identify who is taking the action.

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -99,12 +99,12 @@ export default function todoList (orchestraApi) {
         todoApi.delete(todo)
       }
 
-      todoList.addActionToTodoActivityLog = (todo, action) => {
-        const datetimeUtc = moment.tz(moment(), moment.tz.guess()).utc()
+      todoList.addActionToTodoActivityLog = (todo, action, datetime) => {
+        const activityDatetime = datetime ? datetime : moment.tz(moment(), moment.tz.guess()).utc().format('YYYY-MM-DD HH:mm:ss')
```
marcua: getUTCDateTimeString?

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -116,9 +116,8 @@ export default function todoList (orchestraApi) {
 
       todoList.skipTodo = (todo) => {
         const datetimeUtc = moment.tz(moment(), moment.tz.guess()).utc()
-        todo.skipped_datetime = datetimeUtc.format('YYYY-MM-DD HH:mm')
-        // Not passing todo.skipped_datetime because we want to log datetime at a higher resolution (includes seconds).
-        todoList.addActionToTodoActivityLog(todo, 'skip')
+        todo.skipped_datetime = datetimeUtc.format('YYYY-MM-DD HH:mm:ss')
+        todoList.addActionToTodoActivityLog(todo, 'skip', todo.skipped_datetime)
```
marcua: getUTCDateTimeString

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -100,11 +100,12 @@ export default function todoList (orchestraApi) {
       }
 
       todoList.addActionToTodoActivityLog = (todo, action, datetime) => {
-        const activityDatetime = datetime ? datetime : moment.tz(moment(), moment.tz.guess()).utc().format('YYYY-MM-DD HH:mm:ss')
+        const activityDatetime = datetime || moment.tz(moment(), moment.tz.guess()).utc().format('YYYY-MM-DD HH:mm:ss')
```
marcua: getUTCDateTimeString?

marcua (approved):

PR #475, opened 2018-08-01
Assignment lookup bugfix

marcua:


orchestra/lookups.py
```diff
@@ -75,13 +76,13 @@ def get_query(self, q, request):
 @register('task_assignments')
 class TaskAssignmentLookup(FormatedItemMixin, LookupChannel):
 
-    model = Task
+    model = TaskAssignment
```
marcua, line 79: This was likely a copy/paste error based on `TaskLookup` below. I tested these changes in admin for the two models that use `task_assignments` and the search/save now works :)

thisisdhaas (approved):

PR #476, opened 2018-08-03
Added todolist_templates_to_apply with automatic application

noahpicard:
Added an optional feature todolist templates to steps, which will automatically apply the todolist templates specified to the step when created.


orchestra/models/core/models.py
```diff
@@ -187,39 +250,14 @@ class Step(StepMixin, models.Model):
...
     user_interface = JSONField(default={})
+    todolist_templates_to_apply = models.ManyToManyField(TodoListTemplate,
+                                                         blank=True)
 
     class Meta:
         app_label = 'orchestra'
         unique_together = (('workflow_version', 'slug'),)
 
 
-class Worker(WorkerMixin, models.Model):
```
marcua: nitpick: no need to rearrange these due to dependencies--- you can make the M2M field reference `'TodoListTemplate'` instead of `TodoListTemplate` and be all set :)

orchestra/workflow/load.py
```diff
@@ -206,3 +210,15 @@ def _set_step_dependencies(step, step_data, dependency_attr, dependency_model,
             '{}.{} contains a non-existent slug.'
             .format(step_data['slug'], dependency_attr))
     getattr(step, dependency_attr).set(dependencies)
+
+
+def _set_step_templates(step, step_data, template_attr, template_model,
```
marcua: this looks a lot like _set_step_dependencies. can you call that instead?

orchestra/utils/task_lifecycle.py
```diff
@@ -1137,6 +1138,10 @@ def create_subsequent_tasks(project):
                             status=Task.Status.AWAITING_PROCESSING)
                 task.save()
 
+                # Apply todolist templates to Task
+                for template in task.step.todolist_templates_to_apply.all():
```
marcua, line 1200: add a small test to verify this runs:
* You can model the test off of  the first 3 lines of orchestra/utils/tests/test_task_lifecycle.py:BasicTaskLifeCycleTestCase.test_preassign_workers. Create a test_todolist_templates_to_apply that starts with those lines.
* You can create an empty TodoListTemplateFactory and link the step that will be created when you call `create_subsequent_tasks(project)` to this template.
* `patch` task_lifecycle.add_todolist_template and assert that it was called once with the slug of the `TodoListTemplateFactory`

orchestra/migrations/0079_step_todolist_templates_to_apply.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
marcua: to clean up the migration history, you can revert the two migrations here and create one migration that creates the foreign key-based model.

orchestra/workflow/load.py
noahpicard: Yes! The naming is wrong if we do that- so I'll rename the function also.

orchestra/migrations/0079_step_todolist_templates_to_apply.py
```diff
@@ -0,0 +1,20 @@
...
```
marcua: Awww another migration came in at 0079. Rebase master and recreate?:)

orchestra/tests/helpers/fixtures.py
```diff
@@ -377,13 +379,34 @@ def setup_models(test_case):
         },
     }
 
+    todolist_templates = {
+        'launch-design-checklist': {
```
marcua: more generic names for open source orchestra :)

orchestra/tests/helpers/fixtures.py
```diff
@@ -377,13 +379,34 @@ def setup_models(test_case):
...
     _setup_workers(test_case, workers)
     _setup_projects(test_case, projects)
     _setup_tasks(test_case, tasks)
 
 
+def _setup_todolist_templates(test_case, templates):
+    test_case.todolist_templates = {}
+    for template_idx, template_key in enumerate(templates):
+        template_details = templates[template_key]
+        print(template_details)
```
marcua: remove prints

orchestra/tests/helpers/fixtures.py
```diff
@@ -377,13 +379,34 @@ def setup_models(test_case):
         },
     }
 
+    todolist_templates = {
+        'launch-design-checklist': {
+            'name': 'Launch design checklist',
+            'slug': 'launch-design-checklist'
```
marcua: nitpick: To reduce repetition, you can use the slug from the key in the dict above

orchestra/utils/tests/test_task_lifecycle.py
```diff
@@ -181,9 +181,12 @@ def test_worker_assigned_to_rejected_task(self):
             task__status=Task.Status.POST_REVIEW_PROCESSING)
         self.assertTrue(assignments.exists())
         self.assertTrue(worker_assigned_to_rejected_task(self.workers[4]))
-        with self.assertRaises(TaskAssignmentError):
-            get_new_task_assignment(self.workers[4],
-                                    Task.Status.AWAITING_PROCESSING)
+        with patch('orchestra.utils.task_lifecycle.settings.'
```
marcua, line 184: why was this new patch necessary?

orchestra/utils/tests/test_task_lifecycle.py
```diff
@@ -570,6 +572,19 @@ def test_preassign_workers(self):
         self.assertTrue(
             related_task.is_worker_assigned(self.workers[4]))
 
+    def test_todolist_templates_to_apply(self):
+        project = self.projects['assignment_policy']
+
+        todoListTemplateSlug = 'launch-design-checklist'
```
marcua: no need for the var since you only use the string once :)

orchestra/utils/tests/test_task_lifecycle.py
noahpicard: This test was failing when I ran locally because this setting was false. Insurance against flakiness.

marcua (approved):

PR #479, opened 2018-09-18
Add message_project_team endpoint

jumasheff:


orchestra/project_api/views.py
```diff
@@ -124,3 +125,29 @@ def assign_worker_to_task(request):
...
+        message = data['message']
+        project_id = data['project_id']
+        project = Project.objects.get(id=project_id)
+    except KeyError:
+        text = ('An object with `message` and `project_id` attributes'
+                ' should be supplied')
+        raise BadRequest(text)
+    except Project.DoesNotExist:
+        raise BadRequest('No project for given id')
+    slack = OrchestraSlackService()
```
paopow: Do we need to check that `slack_group_id` is not null?

jumasheff: We should! I'll update the PR.

orchestra/project_api/views.py
```diff
@@ -124,3 +125,35 @@ def assign_worker_to_task(request):
...
+        project_id = data['project_id']
+        project = Project.objects.get(id=project_id)
+    except KeyError:
+        text = ('An object with `message` and `project_id` attributes'
+                ' should be supplied')
+        raise BadRequest(text)
+    except Project.DoesNotExist:
+        raise BadRequest('No project for given id')
+    if project.slack_group_id:
+        slack = OrchestraSlackService()
```
paopow: How about calling `message_expert_slack_group`? https://github.com/b12io/orchestra/blob/fcaede419fe55eb7ee5af6e6feb39e436c2c1fa9/orchestra/utils/notifications.py#L197

paopow (approved):

PR #480, opened 2018-09-24
Add a column on to the list display

paopow:
Add a task column to StaffingRequestInquiry's list display


orchestra/admin.py
```diff
@@ -300,6 +300,7 @@ class StaffingRequestInquiryAdmin(RelatedFieldAdmin, AjaxSelectAdmin):
     })
     list_display = (
         'id', 'communication_preference__worker', 'communication_method',
+        'request__task',
```
marcua: I think that:
* this should point to `request`, and
* the string representation of `request` should be `task - request.cause - request.status`.

If we did that, the dropdown for selecting the appropriate request when you click on a specific inquiry would also be more informative.

marcua (approved):

PR #481, opened 2018-09-26
Add a tags column to dashboard

paopow:
The tags column on the dashboard can be used to show extra information about the project of a task assignment. Users can specify the tags to the project by adding a `tags` field to the `project_data` field.


orchestra/static/orchestra/dashboard/tasktable.html
```diff
@@ -51,6 +51,7 @@ <h3>
                 <th st-sort="project">Project</th>
                 <th st-sort="step">Task</th>
                 <th st-sort="detail">Details</th>
+                <th class="tags-col" st-sort="tag" ng-if="vm.showTagsCol">Tags</th>
```
marcua: st-sort="tag" presumably doesn't work because tag isn't defined yet?

orchestra/models/core/mixins.py
```diff
@@ -174,6 +174,23 @@ def is_entry_level(self):
     def is_reviewer(self):
         return self.assignment_counter > 0
 
+    def _can_view_tag(self, tag, certifications):
+        for tag_certification in tag.get('certificates', []):
```
marcua: verifying tags has no connection to the assignment, so this is more like get_certification_tags.

orchestra/models/core/mixins.py
```diff
@@ -174,6 +174,23 @@ def is_entry_level(self):
...
+            for worker_certification in certifications:
+                cert = worker_certification.certification
+                if (cert.slug == tag_certification['slug'] and
+                        cert.workflow.slug == tag_certification['workflow']):
+                    return True
+        return False
+
+    def get_assignment_tags(self):
+        tags = self.task.project.project_data.get('tags', [])
```
marcua: project_data has no implied schema, so we shouldn't impose/assume one. the solution sketch should look as follows
* Add a field called `task_tags_function` to `Step` (similarly could be done on an assignment). This would be similar to `execution_function` (https://github.com/b12io/orchestra/blob/master/orchestra/models/core/models.py#L142)
* Dynamically load the function and run it on the task (or assignment)
* If you're interested in reading `project_data` (or any other approach), you can do that in a specific `task_tags_function`

Alternatively, given that above I'm saying to move this logic onto tasks from assignments, add a field called `tags` to `Task` (requires a schema migration), specify the JSON schema of that field, and read directly from there instead of running any functions.

marcua (changes requested):
The solution we've got here is pretty different than the one I understood from our sketch. If we go with something like this one, I think we should:
* add a field called `tags` to `Task` (requires a schema migration)
* specify the JSON schema of that field
* and read directly from there (via a TaskSerializer) instead of running any functions

orchestra/models/core/models.py
```diff
@@ -383,6 +383,7 @@ class Status:
     project = models.ForeignKey(
         Project, related_name='tasks', on_delete=models.CASCADE)
     status = models.IntegerField(choices=STATUS_CHOICES)
+    tags = JSONField(default={'tags': []})
```
marcua: ArrayField of charfield :)

orchestra/utils/task_lifecycle.py
```diff
@@ -612,7 +612,9 @@ def tasks_assigned_to_worker(worker):
                 'state': state,
                 'assignment_start_datetime': task_assignment.start_datetime,
                 'next_todo_dict': next_todo_dict,
-                'should_be_active': should_be_active})
+                'should_be_active': should_be_active,
+                'tags': task_assignment.task.tags.get('tags', [])
```
marcua, line 616: if easy, add a test

marcua (approved):
a, but feel free to update tests :)

orchestra/tests/test_dashboard.py
```diff
@@ -79,6 +79,26 @@ def test_status(self):
...
+            worker=self.workers[0])[0].task
+        with self.assertRaises(AttributeError):
+            task.tags = bad_formatted_tags
+            task.tags.save()
+
+        task.tags = valid_tags
+        task.save()
+        response2 = self.clients[0].get('/orchestra/api/interface/dashboard_tasks/')
+        returned2 = load_encoded_json(response2.content)
+        self.assertEqual(len(returned2['tasks'][0]['tags']), 1)
```
marcua: looks good. if you want extra credit, you can 

```
self.assertEqual(returned2['tasks'][0]['tags'][0]['label'], 'foo')
self.assertEqual(returned2['tasks'][0]['tags'][0]['status'], 'default')
```

PR #489, opened 2019-03-07
Optimize get_project_information DB calls

jumasheff:


orchestra/project_api/api.py
```diff
@@ -71,13 +75,16 @@ def get_workflow_steps(workflow_slug, version_slug):
     workflow_version = WorkflowVersion.objects.get(
         slug=version_slug,
         workflow__slug=workflow_slug)
+    steps = Step.objects.prefetch_related(
```
marcua, line 78: very cool! do you need the prefetch_related? i'd imagine that comes for free with the Step

orchestra/project_api/api.py
```diff
@@ -2,6 +2,7 @@
 from collections import defaultdict
```
marcua, line 2: Such a nice PR! Curiously, what's the effect on performance with respect to number of DB queries?

PR #492, opened 2019-04-15
Update node & yarn; Don't consider subset dependency changes to be topology changes

marcua:


orchestra/workflow/load.py
```diff
@@ -1,3 +1,5 @@
+import logging
```
marcua, line 1: The only substantive file changed in this PR

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
           name: Install dependencies
           command: |
             curl -sL https://deb.nodesource.com/setup_7.x | bash - && apt-get install -y nodejs
-            curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.24.4
+            curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
```
marcua, line 14: CI no longer worked because this version was so old 😬

orchestra/static/dist/main.css
```diff
@@ -1919,7 +1919,7 @@ a.logo {
 @charset "UTF-8";
 /*!
  * Pikaday
- * Copyright © 2014 David Bushell | BSD & MIT license | https://dbushell.com/
```
thisisdhaas, line 1922: remind me why this stuff changes? (or why it's even checked in instead of being built at runtime?)

orchestra/workflow/load.py
```diff
@@ -193,11 +197,19 @@ def load_workflow_version(version_data, workflow, force=False):
 def _verify_dependencies_not_updated(step_data, dependency_attr,
                                      old_dependencies):
     new_dependencies = set(step_data.get(dependency_attr, []))
-    if old_dependencies is not None and old_dependencies != new_dependencies:
+    old_set = set(old_dependencies or [])
+    new_set = set(new_dependencies)
+    if old_dependencies is not None and (new_set - old_set):
```
thisisdhaas, line 202: `old_dependencies is not None` feels unnecessary because of your logic on l.200?

thisisdhaas (approved):

orchestra/static/dist/main.css
marcua: We can yarn and check the built code into the repo---could consider cleaning that up int he future :)

PR #531, opened 2019-06-12
Dashboard tmpl todos

paopow:
1. Only count non-templated todo as a next action on the dashboard.
2. Change the date format on the dashboard. If the start by or due on datetime is today, also specify the time. 
image


orchestra/static/orchestra/dashboard/tasktable.directive.es6.js
```diff
@@ -36,6 +36,14 @@ export default function tasktable () {
         vm.collapsed = !vm.collapsed
       }
 
+      vm.getDatetimeFormat = (datetimeString) => {
+        const localTime = datetimeString ? moment.utc(datetimeString).tz(moment.tz.guess()) : null
+        if (localTime && localTime.isSame(new Date(), 'day')) {
+          return '[Today], h:mm a'
```
marcua, line 42: This logic doesn't seem consistent with the screenshot in the PR description
- Today is in brackets---I like it the way it is in your screenshot without them
- There's no case/format for when `!localTime.isSame(new Date(), 'day'))`

LMK if I misread it! :)

paopow: - The brackets are for inserting specific string (in this case, today). See "Escaping characters" here https://momentjs.com/docs/#/displaying/.
- It will use the default provided by the datetime-display directive. https://github.com/b12io/orchestra/blob/master/orchestra/static/orchestra/timing/datetime-display/datetime-display.directive.es6.js#L24

marcua (approved):

PR #576, opened 2020-01-21
Adding deps-upgrade script

dastanko:


git-merge-dep-updates.sh
```diff
@@ -0,0 +1,14 @@
...
+branch="upgrading-dependencies-$(date -I)"
+
+git checkout -b ${branch} > /dev/null
+
+for ((i=0; i < $#; i++))
+{
+    hash=$(git log origin/master..origin/${branches[$i]} --pretty=format:'%h')
+    git cherry-pick ${hash} > /dev/null
+    echo "Picking ${hash} from branch ${branches[$i]}->${branch}"
+}
```
thisisdhaas, line 19: Is it convenient to run `git push -u origin ${branch}` at the end?

git-merge-dep-updates.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
thisisdhaas, line 1: a) Add some explanatory comment / documentation for this command.
b) We'll want this everywhere dependabot runs, right? Is there an easy way to avoid maintaining the same script in 3 repos? e.g., make this script always available on dev1 via a different mechanism than git, or set it up separately as a submodule, or something else?

dastanko: a) Ok.
b) We can put this script in PATH in the dev machine. I think this is a more sustainable way. Are you ok with it?

git-merge-dep-updates.sh
dastanko: Ok, will add this.

git-merge-dep-updates.sh
thisisdhaas: re: b, that sounds good if:
* We still check the file into version control somewhere (maybe https://github.com/b12io/crowdsurfing/tree/master/infrastructure/scripts ?)
* You place it on the path for both dev1 and dev2 after the PR is merged.
* We update the docs for creating new devN machine so we remember to copy it over in the future (or is there a script to which we can add a `cp` command so it happens automatically?)

PR #578, opened 2020-01-23
Update django version to 2.2.9

progremir:
## Changes:
- updated django version from `1.11.10` to `2.2.9`
- changed migrations check command in circle ci
- renamed `MIDDLEWARE_CLASSES` to `MIDDLEWARE`
- replaced `django.core.urlresolvers` with `django.urls`
- added `app_name` definition to `urls.py` files
- updated `djangorestframework` from `3.7.7` to `3.11.0`
- updated `django-filter` from `1.0.4` to `2.2.0`


orchestra/migrations/0060_auto_20160505_1349.py
```diff
@@ -6,6 +6,7 @@
 
 
 class Migration(migrations.Migration):
+    atomic = False
```
marcua, line 9: only required for this migration and not others?

orchestra/tests/test_dashboard.py
```diff
@@ -304,13 +304,13 @@ def test_submit_entry_level_task_assignment(self):
         # user 0 can't submit a task if its submission prerequisites aren't
         # complete
         step = task.step
-        step.submission_depends_on = [
+        step.submission_depends_on.set([
```
marcua, line 307: good catch---does this happen anywhere else in the codebase?

orchestra/views.py
```diff
@@ -336,7 +336,7 @@ class TimeEntryList(generics.ListCreateAPIView):
     permission_classes = (permissions.IsAuthenticated,)
     serializer_class = TimeEntrySerializer
     filter_backends = (filters.DjangoFilterBackend,)
-    filter_class = TimeEntryFilter
+    filterset_class = TimeEntryFilter
```
marcua, line 339: was this just always broken?:) how'd you find it?

docs-requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Packages needed to build docs on https://readthedocs.org
-Django==1.11.10
+Django==2.2.9
```
marcua, line 2: I DMed you an internal manual testing checklist we use -> Happy to run through it live with you, or to test it if you DM me a dev/staging link. Thanks for doing this!

Makefile
```diff
@@ -14,7 +14,7 @@ lint:
 	flake8 . # && isort --check-only --recursive .
 	gulp lint --production
 	!(find . -wholename '**migrations**/*.py' -print0 | xargs -0 grep 'RemoveField\|DeleteModel\|AlterModelTable\|AlterUniqueTogether\|RunSQL\|RunPython\|SeparateDatabaseAndState' | grep -v '# manually-reviewed') || { echo "Migrations need to be manually reviewed!"; exit 1; }
-	!(cd example_project && python3 manage.py makemigrations --dry-run --exit) || { printf "Migrations need to be created. Try: \n\t python3 manage.py makemigrations\n"; exit 1; }
+	cd example_project && python3 manage.py makemigrations --check || { printf "Migrations need to be created. Try: \n\t python3 manage.py makemigrations\n"; exit 1; }
```
marcua: Have you tested that this still fails/stops the build as desired?

progremir: I've tested this on my machine in the following way:
- ran `python3 manage.py makemigrations --dry-run --check || { printf "Migrations need to be created. Try: \n\t python3 manage.py makemigrations\n"; }`. The output was `No changes detected`.
- updated the model field 
- ran `python3 manage.py makemigrations --dry-run --check || { printf "Migrations need to be created. Try: \n\t python3 manage.py makemigrations\n"; }` again. The output was `Migrations need to be created.`

Added `--dry-run` cause it was creating migration files.

orchestra/views.py
progremir: That's because I updated `django-filter`. Since they added django2 support only in v2.0

orchestra/tests/test_dashboard.py
progremir: I didn't find any other places where `ManyToManyField` is set using `=` operation. Usually it's done by `_set_step_relations` method

orchestra/migrations/0060_auto_20160505_1349.py
progremir: Yeah, migrations check was failing with error message like `Cannot rename model atomically in SQLite`.

progremir: I was not sure if we support `SQLite` or force any other dbs :)

PR #596, opened 2020-02-26
Only allow one open staffbot request

paopow:
- Only allow one open staffbot request per task at a time
- Enable search by project description for staffbot related models on the admin interface.

### DIFF
--- a/orchestra/bots/staffbot.py
+++ b/orchestra/bots/staffbot.py
@@ -12,7 +12,7 @@ from orchestra.communication.errors import SlackError
 from orchestra.communication.mail import html_from_plaintext
 from orchestra.communication.mail import send_mail
 from orchestra.communication.slack import format_slack_message
-from orchestra.communication.utils import close_open_staffbot_request
+from orchestra.communication.utils import close_task_open_staffbot_requests
 from orchestra.core.errors import TaskAssignmentError
 from orchestra.core.errors import TaskStatusError
 from orchestra.models import CommunicationPreference
@@ -107,11 +107,9 @@ class StaffBot(BaseBot):
                     'title': 'Error',
                     'text': error_msg
                 }])
+
         # Close open staffbot request
-        if task.has_open_staffing_request():
-            requests = task.staffing_requests.all()
-            for request in requests:
-                close_open_staffbot_request(request)
+        close_task_open_staffbot_requests(task)
 
         StaffBotRequest.objects.create(
                 task=task,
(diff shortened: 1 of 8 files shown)


orchestra/bots/staffbot.py
```diff
@@ -107,12 +108,15 @@ def staff(self, task_id,
                     'text': error_msg
                 }])
 
-        StaffBotRequest.objects.create(
-            task=task,
-            required_role_counter=required_role_counter,
-            request_cause=request_cause)
+        # Close open staffbot request
```
marcua, line 111:
```suggestion
        # Clean up any existing requests before starting a new one.
```

orchestra/bots/staffbot.py
```diff
@@ -170,12 +174,14 @@ def restaff(self, task_id, username,
                     'text': error_msg
                 }])
 
+        # Close open staffbot request
```
marcua, line 177:
```suggestion
        # Clean up any existing requests before starting a new one.
```

orchestra/models/communication/mixins.py
```diff
@@ -41,6 +41,18 @@ def get_request_cause_description(self):
     def get_request_status_description(self):
         return self.Status(self.status).description
 
+    def is_open(self):
+        if self.status == self.Status.PROCESSING.value:
+            return True
+        inquiries = self.inquiries.all()
+        num_responses = 0
+        for inquiry in inquiries:
```
marcua, line 49: rather than looping through all inquiries which involves a lot of round trips to the DB, can you write two queries?
* q1: are there any responses for this request that are winners?
* q2: count the number of responses per inquiry, and count how many inquiries have no responses.

orchestra/communication/utils.py
```diff
@@ -73,3 +73,23 @@ def mark_worker_as_winner(worker, task, required_role_counter,
...
             is_winner=True)
+
+
+def close_open_staffbot_requests(task):
+    requests = task.staffing_requests.all()
+    for request in requests:
+        close_staffbot_request(request)
+
+
+@transaction.atomic
```
marcua, line 84: should we move the atomicity to `close_open_staffbot_requests` so it's all-or-nothing?

orchestra/communication/utils.py
```diff
@@ -73,3 +73,23 @@ def mark_worker_as_winner(worker, task, required_role_counter,
...
+def close_staffbot_request(request):
+    if request.is_open():
+        inquiries = request.inquiries.all()
```
marcua, line 87: similar to the comment below, can we query for all inquiries that don't have responses to reduce round trips?

PR #605, opened 2020-03-04
Only allow one active staffbot request per task

paopow:
- Update the status choices of StaffBotRequest. Add another status that tracks whether the StaffBotRequest is complete. A request is complete when there is a winner response for that request or when there is no winner after all inquiries are responded.
- Close existing staffbot requests when either staff() or restaff() is called.


orchestra/models/communication/models.py
```diff
@@ -102,7 +102,8 @@ class RequestCause(ChoicesEnum):
         RESTAFF = 'restaff'
 
     class Status(ChoicesEnum):
-        PROCESSING = 'processing'
+        SENDING_INQUIRIES = 'sending inquiries'
```
marcua, line 105: enums are append-only, so it's dangerous to insert values before others -> their IDs will get messed up

orchestra/models/communication/models.py
```diff
@@ -102,7 +102,8 @@ class RequestCause(ChoicesEnum):
...
+        WAITING_FOR_RESPONSES = 'waiting for responses'
         COMPLETE = 'complete'
```
marcua: COMPLETE used to mean "i'm done sending inquiries" but now it means "this is no longer an open request" -> what will that mean for existing COMPLETE requests that are still open? they will presumably disappear from the available tasks list.

you could consider renaming COMPLETE to FINISHED_SENDING_INQUIRIES (perhaps that's a better name for WAITING_FOR_RESPONSES) and then starting off with no COMPLETE requests in the DB?

orchestra/models/communication/models.py
marcua: depending on the comment below, you might be fine, as id = 1 (used to be COMPLETE) will now be WAITING_FOR_RESPONSES. we can discuss after you think both comments through.

orchestra/communication/staffing.py
```diff
@@ -156,7 +156,7 @@ def _send_request_inquiries(staffbot, request, worker_batch_size,
             request.task.project.slack_group_id,
             ('All staffing requests for task {} have been sent!'
              .format(request.task)))
-        request.status = StaffBotRequest.Status.COMPLETE.value
+        request.status = StaffBotRequest.Status.WAITING_FOR_RESPONSES.value
```
marcua: Would a better name be FINISHED_SENDING_INQUIRIES?

orchestra/communication/staffing.py
```diff
@@ -192,6 +192,7 @@ def get_available_requests(worker):
     remaining_requests = (
         StaffBotRequest.objects
         .filter(inquiries__communication_preference__worker=worker)
+        .exclude(status=StaffBotRequest.Status.COMPLETE.value)
```
marcua: sanity-checking: is this function the one that's used for the available tasks page? have we manually tested the available tasks page to ensure that no existing requests disappear after this PR, and that requests do disappear after being assigned or completed?

orchestra/communication/tests/test_models.py
```diff
@@ -82,3 +83,57 @@ class StaffingRequestInquiryTestCase(OrchestraModelTestCase):
...
     model = StaffingResponseFactory
+
+    def test_mark_winner(self):
+        request = StaffBotRequestFactory()
+        inquiry = StaffingRequestInquiryFactory(request=request)
+        self.assertEqual(
+            request.status,
+            StaffBotRequest.Status.SENDING_INQUIRIES.value)
+
+        # The staff bot request status should be complete when one
```
marcua: i agree with the comment, but don't understand how the code works -> where in this test are we calling something like `mark_worker_as_winner` or something like that?

orchestra/communication/utils.py
```diff
@@ -27,7 +27,8 @@ def mark_worker_as_winner(worker, task, required_role_counter,
     else:
```
marcua, line 26: do we have to check that the staffing request is still valid? perhaps this can be done in the filter on lines 17-19?

on a related note: is there a test that ensures that if you're the first winner of a request, but that request is already closed, you don't win?

orchestra/communication/utils.py
```diff
@@ -27,7 +27,8 @@ def mark_worker_as_winner(worker, task, required_role_counter,
...
         staffbot_request = staffbot_request.first()
 
-    staffbot_request.status = StaffBotRequest.Status.COMPLETE.value
+    staffbot_request.status = (
```
marcua: shouldn't a winner mark the task as complete?

orchestra/communication/utils.py
```diff
@@ -27,7 +27,8 @@ def mark_worker_as_winner(worker, task, required_role_counter,
...
+        StaffBotRequest.Status.WAITING_FOR_RESPONSES.value)
     staffbot_request.save()
 
     # Mark everyone else as non-winner
```
marcua: do we need to do this anymore if we mark the winner and mark the task COMPLETE?

orchestra/models/communication/mixins.py
```diff
@@ -53,6 +53,7 @@ class StaffingRequestInquiryMixin(object):
 
     def __str__(self):
         return '{} - {}'.format(
+            self.id,
```
marcua: do you have to change the format string?

orchestra/models/communication/mixins.py
```diff
@@ -66,3 +67,35 @@ def __str__(self):
             self.is_available,
             self.is_winner
         )
+
+    def _mark_staffbot_request_complete(self, request):
+        from orchestra.models import StaffBotRequest
+        request.status = StaffBotRequest.Status.COMPLETE.value
+        request.save()
+
+    def save(self, *args, **kwargs):
```
marcua: not a strong opinion: would this this logic more cleanly live in whatever functions get called when a worker responds accept/ignore rather than in this save method?

orchestra/models/communication/mixins.py
```diff
@@ -66,3 +67,35 @@ def __str__(self):
...
+            responded_inquiries = inquiries.filter(
+                responses__isnull=False).distinct()
+            num_responded_workers = len(
+                set(responded_inquiries.values_list(
+                    'communication_preference__worker__id', flat=True)
+                    )
+            )
+
+            if num_responded_workers >= num_inquired_workers:
+                self._mark_staffbot_request_complete(request)
```
marcua: this is nice cleanup. do we need a data migration to do this for all old projects, or will some other logic mark them as complete eventually?

orchestra/models/communication/mixins.py
```diff
@@ -66,3 +67,35 @@ def __str__(self):
...
+
+            responded_inquiries = inquiries.filter(
+                responses__isnull=False).distinct()
+            num_responded_workers = len(
+                set(responded_inquiries.values_list(
+                    'communication_preference__worker__id', flat=True)
+                    )
+            )
+
+            if num_responded_workers >= num_inquired_workers:
```
marcua: this is nice cleanup. do we need a data migration to do this for all old projects, or will some other logic mark them as complete eventually?

orchestra/communication/tests/test_models.py
marcua: i see -> this happens in the new `save` method. see my comment on the save method asking if there's a cleaner/more explicit place in the API to put it

orchestra/models/communication/models.py
paopow: The previous COMPLETE status means that all the requests are sent out. In this case, WAITING_TO_RESPONSES means the same thing just more specific. The new enum COMPLETE is used to mark when there is the request is considered "closed" which is when 1) there is a winner for the task, or 2) all qualified workers declined the tasks.

orchestra/models/communication/mixins.py
paopow: Nope. It was for debugging. I will remove this.

orchestra/models/communication/models.py
paopow: > enums are append-only, so it's dangerous to insert values before others -> their IDs will get messed up

I think it should be fine since we are changing just the labels and not the semantic meaning of the enums. (Am I wrong?)

orchestra/communication/utils.py
paopow: you're right

orchestra/communication/staffing.py
paopow: Yup.

orchestra/communication/utils.py
paopow: Yup

orchestra/bots/tests/test_staffbot.py
```diff
@@ -172,6 +172,37 @@ def test_staff_command_errors(self):
                          .format(task.id,
                                  'Status incompatible with new assignment'))
 
+    @patch('orchestra.bots.staffbot.send_mail')
+    @patch('orchestra.bots.staffbot.message_experts_slack_group')
+    @patch('orchestra.bots.staffbot.StaffBot._send_staffing_request_by_slack')
+    def test_staff_close_requests(self, mock_slack,
+                                  mock_experts_slack, mock_mail):
+        """
+        Test that existing staffbot requests for a task is close when
```
marcua:
```suggestion
        Test that existing staffbot requests for a task is closed when
```

orchestra/bots/tests/test_staffbot.py
```diff
@@ -240,6 +271,41 @@ def test_restaff_command_errors(self):
                          (bot.task_assignment_does_not_exist_error
                           .format(worker.user.username, task.id)))
 
+    @patch('orchestra.bots.staffbot.send_mail')
+    @patch('orchestra.bots.staffbot.message_experts_slack_group')
+    @patch('orchestra.bots.staffbot.StaffBot._send_staffing_request_by_slack')
+    def test_restaff_close_requests(self, mock_slack,
+                                    mock_experts_slack, mock_mail):
+        """
+        Test that existing staffbot requests for a task is close when
```
marcua:
```suggestion
        Test that existing staffbot requests for a task is closed when
```

orchestra/communication/staffing.py
```diff
@@ -59,17 +59,16 @@ def handle_staffing_response(worker, staffing_request_inquiry_id,
                 'Cannot reject after accepting the task')
 
         response.is_available = is_available
+        response.save()
```
marcua: curiously, why did we need to add this and didn't need it before?

orchestra/communication/tests/test_staffing.py
```diff
@@ -433,13 +441,30 @@ def test_mark_worker_as_winner(self):
...
+            request__task__step__is_human=True,
+            request=self.staffing_request_inquiry.request
+        )
+        mark_worker_as_winner(worker2,
+                              inquiry2.request.task,
+                              0,
+                              inquiry2)
+        self.assertEqual(inquiry2.responses.count(), 0)
+
+        # A winner can't declined their accpeted task
```
marcua:
```suggestion
        # A winner can't decline their accepted task
```

orchestra/bots/staffbot.py
```diff
@@ -107,6 +108,7 @@ def staff(self, task_id,
                     'text': error_msg
                 }])
 
+        close_open_staffbot_requests(task)
```
marcua, line 111: these calls are good. we should also call `close_open_staffbot_requests` if a worker is manually assigned to a task (e.g., through project management or if they pick up a task from their dashboard). once you add that call, you can also clean up code that previously created a "fake" response from the manually assigned worker.

orchestra/migrations/0083_update_staffbot_request_status_choices.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 2.2.9 on 2020-03-30 19:21
+
+from django.db import migrations, models
+
+def migrate_request_status(apps, schema_editor):
```
marcua, line 5: how long does the migration take to run?

orchestra/migrations/0083_update_staffbot_request_status_choices.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 2.2.9 on 2020-03-30 19:21
+
+from django.db import migrations, models
+
```
marcua, line 4: (you can DM me this for privacy reasons) Manual testing
* create an up-to-date production snapshot
* Report on count of request statuses.
* Record the Available Tasks screen of a user with lots of available tasks (maybe a CSS, a copywriter, and a designer)
* Run the migration
* Look at the available tasks lists. They should look the same.
* Report on count of request statuses.

orchestra/migrations/0083_update_staffbot_request_status_choices.py
```diff
@@ -0,0 +1,56 @@
...
+    requests = (StaffBotRequest.objects.all()
+                .exclude(status=CLOSED))
+    for request in requests:
+        responses = StaffingResponse.objects.filter(
+                        request_inquiry__request=request)
+        has_winner = responses.filter(is_winner=True).exists()
+        if has_winner:
+            request.status = CLOSED
+        elif request.status == DONE_SENDING_INQUIRIES:
+            inquiries = (
```
marcua, line 22: sanity-checking that this is a copy/paste of the latest iteration of your `check_responses_complete` code

orchestra/bots/staffbot.py
marcua: (apologies if you did this and i didn't see it somewhere)

orchestra/migrations/0083_update_staffbot_request_status_choices.py
paopow: About 40s.

orchestra/migrations/0083_update_staffbot_request_status_choices.py
paopow: Yes, it is.

orchestra/bots/staffbot.py
paopow: It is already taken care of in `mark_worker_as_winner` which is called by `assign_task` and `reassign_task`. But you give me an idea of how to make `mark_worker_as_winner` a little cleaner.

orchestra/communication/staffing.py
paopow: Technically we don't need it. I just added this for clarity.

The reason we didn't have to save it was because if `is_available` is true, then we call `assign_task` or `reassign_task` and those function set `is_available` to true together with `is_winner`.  When `is_available` is false, the default value is already false and there is no need to save it. 

We could remove it since it is doing nothing but the current code is not super clear either.

orchestra/migrations/0083_update_staffbot_request_status_choices.py
paopow: Responded in DM.

orchestra/communication/utils.py
```diff
@@ -28,9 +28,7 @@ def mark_worker_as_winner(worker, task, required_role_counter,
     else:
         staffbot_request = staffbot_request.first()
 
-    staffbot_request.status = (
```
marcua: With this, do we need lines 46-74?

orchestra/communication/staffing.py
marcua: Very much your call, as you're closer to the code. Sometimes saving twice bites you beyond performance reasons if you safe an extraneous field you weren't ready to save yet, but whatever you decide :).

paopow: OK. I will remove it but will leave a comment for future references.

orchestra/communication/tests/test_staffing.py
```diff
@@ -433,22 +441,25 @@ def test_mark_worker_as_winner(self):
...
-                         StaffBotRequest.Status.COMPLETE.value)
+                         StaffBotRequest.Status.CLOSED.value)
         self.assertEqual(
-            self.staffing_request_inquiry.responses.all().count(), 1)
-
-        response = self.staffing_request_inquiry.responses.first()
-        self.assertTrue(response.is_winner)
+            self.staffing_request_inquiry.responses.all().count(), 0)
 
-        response.is_winner = False
```
paopow, line 443: We already catch this case in line 98.

marcua (approved):
woooo!

PR #621, opened 2020-06-05
Allows re-loading a workflow version with --force if it has new steps.

thisisdhaas:
This PR makes it possible to add steps to the leaves of a workflow by re-loading it with the `--force` option.
* Previously, any difference in steps between the old and new versions would result in an error.
* This PR still prevents removing steps from the old version, or inserting steps before other steps, as that could break existing projects.
* The logic for why this change is safe is as follows:
   * Since adding a new step doesn't change the `created_depends_on` or `submission_depends_on` properties of any other step, existing projects can't be prevented from proceeding.
   * Since new steps can't be inserted into the topology as antecedents of another step, existing projects can't be prevented from completing.
   * For existing projects, calling `create_subsequent_tasks()` after adding Steps to the workflow might result in new tasks being created, but they won't be created by default upon loading the new steps.

### DIFF
--- a/orchestra/workflow/load.py
+++ b/orchestra/workflow/load.py
@@ -128,11 +128,11 @@ def load_workflow_version(version_data, workflow, force=False):
             .filter(workflow_version=version)
             .values_list('slug', flat=True)
         )
-        if new_step_slugs != old_step_slugs:
-            raise WorkflowError('Even with --force, you cannot change the '
-                                'steps of a workflow. Drop and recreate the '
-                                'database to reset, or create a new version '
-                                'for your workflow.')
+        if old_step_slugs - new_step_slugs:
+            raise WorkflowError(
+                'Even with --force, you cannot remove steps from a workflow.'
+                'Drop and recreate the database to reset, or create a new '
+                'version for your workflow.')
 
     # Create or update the version steps.
     old_creation_dependencies = {}
--- a/orchestra/tests/helpers/workflow.py
+++ b/orchestra/tests/helpers/workflow.py
@@ -876,11 +876,12 @@ def assert_test_dir_v1_loaded(test_case, extra_step=False):
                     ],
                 },
             })
-        test_case.assertEqual(set(step4.required_certifications.all()),
-                              set(Certification.objects.filter(
-                                  slug__in=['certification1', 'certification2'],
-                                  workflow__slug='test_dir',
-                              )))
+        test_case.assertEqual(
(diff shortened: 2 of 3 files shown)


orchestra/workflow/load.py
```diff
@@ -128,11 +128,11 @@ def load_workflow_version(version_data, workflow, force=False):
             .filter(workflow_version=version)
             .values_list('slug', flat=True)
         )
-        if new_step_slugs != old_step_slugs:
-            raise WorkflowError('Even with --force, you cannot change the '
-                                'steps of a workflow. Drop and recreate the '
-                                'database to reset, or create a new version '
-                                'for your workflow.')
+        if old_step_slugs - new_step_slugs:
```
thisisdhaas, line 131: the critical change: only checks that `new_step_slugs` isn't missing anything that was in `old_step_slugs`.

orchestra/workflow/load.py
```diff
@@ -128,11 +128,11 @@ def load_workflow_version(version_data, workflow, force=False):
...
+            raise WorkflowError(
+                'Even with --force, you cannot remove steps from a workflow.'
+                'Drop and recreate the database to reset, or create a new '
```
marcua, line 134: I love that we lead with `Drop and recreate the database to reset` rather than `looks like you changed something, bump the version` :)

orchestra/workflow/load.py
marcua: In the PR description, you made an argument that we should only do this for leaf nodes, but here theoretically i can insert a node in between two others, no? Or does some logic up above prevent that?

thisisdhaas: Subsequent logic (l.184) checks that no dependencies have changed, so you can't update the step that would come 'after' to say `creation_depends_on.append(my_new_step)`.

If you're thinking about the submission dependency graph, you can insert a node into the graph by saying that your new step has `submission_depends_on = some_existing_step`. However, that doesn't feel risky because it only affects whether or not you can submit your new task, it doesn't affect other tasks. (This script doesn't even verify the submission constraint topology, so you can mess with it at will anyways :-) ).

orchestra/workflow/load.py
thisisdhaas: I guess it's a helpful message if you're developing your very first workflow, where dropping the DB is probably the right move.

marcua (approved):

PR #663, opened 2020-08-12
Task unification: add fields to todo

jumasheff:


orchestra/models/core/models.py
```diff
@@ -659,9 +659,16 @@ class Todo(TodoMixin, BaseModel):
     class Meta:
```
marcua, line 660: Migration plz :)

orchestra/models/core/models.py
```diff
@@ -672,6 +679,8 @@ class Meta:
         TodoListTemplate, null=True, blank=True, related_name='template',
         on_delete=models.SET_NULL)
     activity_log = JSONField(default={'actions': []})
+    delivery_status = models.IntegerField(null=True, blank=True)
```
marcua:
```suggestion
    status = models.IntegerField(null=True, blank=True)
```

marcua: Add a ChoicesEnum for the three statuses and `choices=Status.choices()`

orchestra/models/core/models.py
```diff
@@ -659,9 +660,21 @@ class Todo(TodoMixin, BaseModel):
...
 
-    task = models.ForeignKey(
-        Task, related_name='todos', on_delete=models.CASCADE)
+    class Status(ChoicesEnum):
+        PENDING = 'pending'
+        COMPLETED = 'completed'
+        DECLINED = 'declined'
+
+    task = models.ForeignKey(Task, null=True, blank=True,
+                             related_name='todos', on_delete=models.CASCADE)
```
marcua:
```suggestion
                             related_name='todos', on_delete=models.SET_NULL)
```

marcua: Now that task is less prominent, we should make it so deleting a task doesn't delete a todo

marcua (approved):

PR #665, opened 2020-08-16
Todo list template export/import via Google Sheets

marcua:
This PR allows us to export and import to-do list templates as Google Sheets. Previously, to edit a template, you had to edit some pretty gnarly JSON. Specifically, this PR introduces
* Convenience functions for exporting to and importing CSVs from Google Sheets
* Admin object actions to import and export to-do list templates
* Logic to import and export the gnarly to-do list template JSON to tabular format
* A model for tracking to-do list template import records for provenance purposes.

Here's a B12-anonymized demo: https://www.loom.com/share/5bf1483242e84efb8c86d7afa4ccda96

Note: the PR is pretty big. I'm happy to break it apart if it helps. The good news is that almost half of it is tests, so...profit?


orchestra/admin.py
```diff
@@ -192,8 +197,22 @@ class TodoQAAdmin(AjaxSelectAdmin):
     search_fields = ('todo__title', 'comment',)
```
marcua, line 199: This file's changes largely add an admin entry for the import records and buttons for triggering import/export

orchestra/google_apps/service.py
```diff
@@ -27,26 +27,27 @@ def _create_drive_service(self, google_p12_path,
...
         """Insert new file.
 
         Args:
           service: Drive API service instance.
           title: Title of the file to insert, including the extension.
           description: Description of the file to insert.
           parent_id: Parent folder's ID.
-          mime_type: MIME type of the file to insert.
+          file_mime_type: MIME type of the file to insert.
+          google_mime_type: MIME type of the file in Google Drive.
```
marcua, line 40: Turns out we were passing the same MIME type for both the file and the Google Drive doc type, which worked in the two existing use cases in convenience.py above (refactored appropriately) but not for our use case of uploading a CSV and having it be converted to a Google Sheet.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
```diff
@@ -0,0 +1,47 @@
+{% extends "admin/base_site.html" %}
```
marcua, line 1: Haven't written one of these in a while. Internal tooling FTW! :)

orchestra/tests/test_todos.py
```diff
@@ -1,13 +1,18 @@
+import copy
```
marcua, line 1: This update is an eyesore. I'm sorry. If it helps to get a walkthrough, LMK.

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,100 @@
+import csv
```
marcua, line 1: This file contains the only interesting business logic in the PR: Import and export!

orchestra/tests/test_todos.py
```diff
@@ -602,3 +631,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+        self.import_url_name = (
+            'orchestra:todos:import_todo_list_template_from_spreadsheet')
+        self.template_admin_name = 'admin:orchestra_todolisttemplate_change'
+        # Object actions don't appear to have a `reverse` pattern, so
+        # we explicitly spell out the URL.
+        self.export_url_pattern = (
+            '/vip/orchestra/todolisttemplate/{}/actions/export_spreadsheet/')
+
+    @patch('orchestra.todos.import_export._upload_csv_to_google')
+    def test_export_spreadsheet(self, mock_upload):
```
marcua, line 722: This test succeeds in dev but fails in CI. Will investigate.

orchestra/tests/test_todos.py
```diff
@@ -602,3 +632,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+        mock_upload.return_value = 'https://redirect.com/the_spreadsheet'
+        response = self.request_client.get(
+            self.export_url_pattern.format(self.full_todo_list_template.id))
+        self.assertEqual(response.status_code, 302)
```
marcua, line 750: This fails with a 404 in CI but succeeds in dev. I suspect somehow `django-object-actions` isn't being registered/installed properly, despite being installed in the `Install dependencies` step. I'll keep poking around, but any ideas are welcome. The other tests pass in both environments.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
```diff
@@ -0,0 +1,47 @@
...
+{% block extrastyle %}
+<link rel="stylesheet" href="{% static 'admin/css/forms.css' %}">
+{% include "admin/inc/extrastyle.html" %}
+{% endblock %}
+
+{% block content %}
+<div id="content-main">
+    <h1>Import todo list template from a Google Sheet</h1>
+
+    <form action="{{request.path}}" method="post" id="churn_form">
```
jumasheff: `churn_form`?

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
```diff
@@ -0,0 +1,47 @@
...
+        {% csrf_token %}
+        <div id="form_{{ form.form_id }}" class="churn-form">
```
jumasheff: `churn-form`? Do we have a reason why we use this class?

orchestra/tests/test_todos.py
```diff
@@ -602,3 +632,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+            },
+            todos={}
+        )
+
+        self.import_url_name = (
+            'orchestra:todos:import_todo_list_template_from_spreadsheet')
+        self.template_admin_name = 'admin:orchestra_todolisttemplate_change'
+        # Object actions don't appear to have a `reverse` pattern, so
+        # we explicitly spell out the URL.
+        self.export_url_pattern = (
```
jumasheff: `reverse("admin:<app_label>_<model_name>_actions", args=[1, "foo"])`

I'd try: `reverse("admin:orchestra_todolisttemplate_actions", args=[123, "export_spreadsheet"])`, but not 100% sure if works.

orchestra/admin.py
```diff
@@ -204,6 +223,19 @@ class TodoListTemplateAdmin(AjaxSelectAdmin):
...
+    def export_spreadsheet(self, request, todo_list_template):
+        return HttpResponseRedirect(export_to_spreadsheet(todo_list_template))
+    export_spreadsheet.attrs = {'target': '_blank'}
+    export_spreadsheet.short_description = 'Export to spreadsheet'
+    export_spreadsheet.label = 'Export to spreadsheet'
+
+    def import_spreadsheet(self, request, todo_list_template):
+        return redirect(
+            'orchestra:todos:import_todo_list_template_from_spreadsheet',
+            pk=todo_list_template.id)
```
thisisdhaas, line 235: Making sure I get it, 'import' only operates on an existing TodoListTemplate, right? So there's no way to 'import + create', only 'import + overwrite'? Doesn't feel like a big issue, but I guess the future workflow for creating a new template will involve:
* Create an empty TodoListTemplate
* Create a spreadsheet w/ the template contents
* Import the spreadsheet into the empty template

orchestra/core/errors.py
```diff
@@ -72,3 +72,7 @@ class TimerError(Exception):
 
 class ProjectStatusError(Exception):
     pass
+
+
+class SpreadsheetImportError(Exception):
```
thisisdhaas: Just quibbling, but as written this doesn't make sense to me in core/errors.py (I think of import errors as things like 'error pulling data from an external service', which aren't Orchestra's problem).

It would make sense to me if you called it 'TodoListTemplateValidationError', which is an Orchestra concern.

orchestra/google_apps/convenience.py
```diff
@@ -217,6 +223,16 @@ def upload_file(parent_id, file_path, title, description, mimetype):
...
         file_path
     )
     return file_metadata
+
+
+def get_google_spreadsheet_as_csv(gdoc_url, reader=csv.DictReader):
+    matches = GSPREAD_RE.match(gdoc_url)
+    assert matches is not None
+    export_url = GSPREAD_EXPORT_URL.format(matches.group(1))
+    response = requests.get(export_url)
```
thisisdhaas, line 238: Feels funny to do this with `requests.get()`, why not use the Drive API directly? https://developers.google.com/drive/api/v3/reference/files/export looks pretty straightforward.

orchestra/models/core/models.py
```diff
@@ -592,7 +593,7 @@ class PayRate(PayRateMixin, models.Model):
     end_date = models.DateField(null=True, blank=True)
 
 
-class TodoListTemplate (TodoListTemplateMixin, BaseModel):
+class TodoListTemplate(TodoListTemplateMixin, BaseModel):
```
thisisdhaas, line 596: 👀

orchestra/admin.py
```diff
@@ -192,8 +197,22 @@ class TodoQAAdmin(AjaxSelectAdmin):
...
+        'todo_list_template__description',
+        'import_url'
+    )
+    ordering = ('-created_at',)
+
+
 @admin.register(TodoListTemplate)
-class TodoListTemplateAdmin(AjaxSelectAdmin):
+class TodoListTemplateAdmin(DjangoObjectActions, AjaxSelectAdmin):
+    change_actions = ('export_spreadsheet', 'import_spreadsheet')
```
thisisdhaas, line 215: what is the access control on these actions? Do only superusers have access by default?

orchestra/models/core/models.py
```diff
@@ -625,6 +626,31 @@ class Meta:
     conditional_property_function = JSONField(default={})
 
 
+class TodoListTemplateImportRecord(
+        TodoListTemplateImportRecordMixin, BaseModel):
```
thisisdhaas, line 630: `BaseModel` gives this a created_at, right?

thisisdhaas: (if so, might be worth adding to the docstring just to be clear that that's the timestamp that tells us when the import happened)

orchestra/models/core/models.py
```diff
@@ -625,6 +626,31 @@ class Meta:
...
+    class Meta:
+        app_label = 'orchestra'
+
+    todo_list_template = models.ForeignKey(
+        TodoListTemplate, null=True, blank=True,
+        on_delete=models.CASCADE)
+    importer = models.ForeignKey(
+        settings.AUTH_USER_MODEL, null=True, blank=True,
+        on_delete=models.SET_NULL)
+    import_url = models.URLField(null=True, blank=True)
```
thisisdhaas, line 654: Is the idea that we should be able to reverse-engineer the contents that were imported by opening the source spreadsheet and looking at the document history as of the created_at timestamp?

(an alternate approach is to store the raw spreadsheet contents as a bytes field, which is obviously pretty heavyweight as a logging mechanism)

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
```diff
@@ -0,0 +1,47 @@
...
+          <div class="aligned">
+            <div class="form-row" id="{{ field.id_for_label }}_row">
+              <div {% if field.errors %}class="field_error"{% endif %}>
+                <span>{{ field.label_tag }}</span>
+                <div class="field">
+                  {{ field }}
+                  <wbr>
+                  <div class="help-text">{{ field.help_text }}</div>
+                  {% for err in field.errors %}
+                  <span class="error_message">{{ err }}</span>
```
thisisdhaas, line 21: Glancing at your loom video, we could make the error stand out a bit more clearly by:
* Turning the error text red
* Adding some bottom margin to separate it from the 'Submit' button.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
```diff
@@ -0,0 +1,47 @@
...
+        <div id="form_{{ form.form_id }}" class="churn-form">
+          {% for field in form %}
+          <div class="aligned">
+            <div class="form-row" id="{{ field.id_for_label }}_row">
+              <div {% if field.errors %}class="field_error"{% endif %}>
+                <span>{{ field.label_tag }}</span>
+                <div class="field">
+                  {{ field }}
+                  <wbr>
+                  <div class="help-text">{{ field.help_text }}</div>
```
thisisdhaas: Do we have help text? (Is this copy-pasta from somewhere?)

thisisdhaas: Worth refactoring it as a generic `import_spreadheet_form.html` so we can use it with future Spreadsheet import admin actions? I don't actually see anything todo-list specific, so perhaps "refactor" = "rename".

orchestra/tests/test_todos.py
```diff
@@ -22,6 +28,30 @@
 from orchestra.utils.load_json import load_encoded_json
 
 
+GOOD_CSV_TEXT = """Remove if,Skip if
```
thisisdhaas: to reduce the eyesore-ishness, can you move these module-level vars plus the bigger json blocks from specific tests into a `fixtures.py`?

orchestra/tests/test_todos.py
```diff
@@ -22,6 +28,30 @@
...
+[],[],the root
+[],[],,todo parent 1
+[],"[{""prop"": {""value"": true, ""operator"": ""==""}}]",,,todo child 1-1
+"[{""prop"": {""value"": true, ""operator"": ""==""}}]",[],,todo parent 2
+[],[],,,todo child 2-1
+[],[],,,,todo child 2-1-1
+[],[],,,todo child 2-2
+"""
+
+BAD_HEADER_CSV_TEXT = """,Remove if,Skip if
```
thisisdhaas: bad just because of the extra column up front?

orchestra/tests/test_todos.py
```diff
@@ -602,3 +632,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+        self.worker = Worker.objects.get(user__username='test_user_6')
+        self.worker.user.is_staff = True
+        self.worker.user.save()
+        self.request_client.login(username=self.worker.user.username,
+                                  password='defaultpassword')
+        self.full_todo_list_template = TodoListTemplateFactory(
+            slug='full-template-slug',
+            name='Full template name',
+            description='Full template description',
+            conditional_property_function={
```
thisisdhaas, line 720: only asking because I saw this here and don't know TodoListTemplates too well: What's the condition_property_function, and is it something we need to import/export?

orchestra/tests/test_todos.py
thisisdhaas: This matches my expectation (where `123` in Murat's example above is the TodoListTemplate id).

orchestra/tests/test_todos.py
thisisdhaas: Hm, no obvious answers come to mind. One idea to help debug: verify that admin urls are available at all, by hitting a non-django-object-actions admin endpoint (like the `_change` endpoint for the same TodoListTemplate) in CI and looking for a 200.

orchestra/tests/test_todos.py
```diff
@@ -602,3 +632,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+        self.assertEqual(response.url, 'https://redirect.com/the_spreadsheet')
+        self.assertEqual(mock_upload.call_count, 1)
+        title_prefix = (mock_upload.call_args[0][0]
+                        [:len(self.full_todo_list_template.name) + 2])
+        self.assertEqual(
+            title_prefix,
+            '{} -'.format(self.full_todo_list_template.name))
+        with open(mock_upload.call_args[0][1].name, 'r') as exported_file:
+            lines = ''.join(exported_file.readlines())
+            # The JSON encoding of properties is nondeterministic in order.
```
thisisdhaas, line 760: I guess you could `json.loads()` and compare the dicts directly, but I don't feel strongly.

orchestra/tests/test_todos.py
```diff
@@ -602,3 +632,241 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+        # Before import, the empty_todo_list_template should have no
+        # template to-dos. After import, it should have the same
+        # to-dos as full_todo_list_template (the CSV we load from is
+        # equivalent to full_todo_list_template's to-dos).
+        self.empty_todo_list_template.refresh_from_db()
+        self.assertEqual(self.empty_todo_list_template.todos, {})
+        self.assertEqual(
+            TodoListTemplateImportRecord.objects.all().count(),
+            0)
+        response = self.request_client.post(
```
thisisdhaas, line 800: I'm surprised this doesn't fail CSRF checks—does Django disable those in tests?

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+    sheet = service.insert_file(
+        spreadsheet_name,
+        '',
+        settings.ORCHESTRA_TODO_LIST_TEMPLATE_EXPORT_GDRIVE_FOLDER,
+        'text/csv',
+        'application/vnd.google-apps.spreadsheet',
+        file.name
+    )
+    service.add_permission(sheet['id'], write_with_link_permission)
+    return 'https://docs.google.com/spreadsheets/d/{}'.format(sheet['id'])
```
thisisdhaas: Is 'sheet' the JSON response from the Google Drive API? I'd be surprised if it doesn't have a URL to the file on it in addition to the id.

Update: in the docs, maybe it's 'selfLink' or 'alternateLink'? https://developers.google.com/drive/api/v2/reference/files#resource

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+from orchestra.google_apps.service import Service
+from orchestra.google_apps.convenience import get_google_spreadsheet_as_csv
+from orchestra.models import TodoListTemplateImportRecord
+
+
+REMOVE_IF_HEADER = 'Remove if'
+SKIP_IF_HEADER = 'Skip if'
+
+
+def _write_template_rows(writer, todo, depth):
```
thisisdhaas, line 20: Took me a minute to get this, perhaps worth a docstring like "a row looks like `{remove_if_json}, {skip_if_json}, {empty_columns if needed so that children are nested underneath parents}, {todo_item_description}`"

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def export_to_spreadsheet(todo_list_template):
+    """
+    Recursively descend down the to-do tree of a `todo_list_template`
+    and add a row per to-do to a CSV, uploading it to a Google Sheet
+    when done. The first two columns of the sheet capture json-encoded
+    remove_if and skip_if logic. After that, the text in the `i`'th
+    column represents a to-do's title that is `i` levels nested in the
+    to-do tree.
```
thisisdhaas, line 61: oh I see this was the docstring I wanted above :-). Maybe above add "see `export_to_spreadsheet()` for a description of each row's format."

orchestra/google_apps/convenience.py
```diff
@@ -217,6 +223,16 @@ def upload_file(parent_id, file_path, title, description, mimetype):
...
         mimetype,
+        mimetype,
         file_path
     )
     return file_metadata
+
+
+def get_google_spreadsheet_as_csv(gdoc_url, reader=csv.DictReader):
+    matches = GSPREAD_RE.match(gdoc_url)
+    assert matches is not None
```
thisisdhaas: raise a nicer error? (This is what shows up in the interface if you typo a URL or try to import a non-google-spreadsheets URL, right?)

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+SKIP_IF_HEADER = 'Skip if'
+
+
+def _write_template_rows(writer, todo, depth):
+    writer.writerow(
+        [json.dumps(todo.get('remove_if', [])),
+         json.dumps(todo.get('skip_if', []))] +
+        ([''] * depth) +
+        [todo.get('description', '')])
+    # `reversed` iteration because the JSON-serialized order of
```
thisisdhaas, line 32: FWIW I still don't understand why we did this.

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+    and add a row per to-do to a CSV, uploading it to a Google Sheet
+    when done. The first two columns of the sheet capture json-encoded
+    remove_if and skip_if logic. After that, the text in the `i`'th
+    column represents a to-do's title that is `i` levels nested in the
+    to-do tree.
+    """
+    with NamedTemporaryFile(mode='w+', delete=False) as file:
+        writer = csv.writer(file)
+        writer.writerow([REMOVE_IF_HEADER, SKIP_IF_HEADER])
+        _write_template_rows(writer, todo_list_template.todos, 0)
```
thisisdhaas, line 66: Is there always a single root for a todo_list_template's items? `_write_template_rows()` seems to expect a single `todo`, whereas this function passes `todo_list_template.todos`.

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+                            for columnindex, text in enumerate(row[2:])
+                            if text]
+        if len(nonempty_columns) == 0:
+            continue
+        elif len(nonempty_columns) > 1:
+            raise SpreadsheetImportError(
+                'More than one text entry in row {}: {}'.format(
+                    rowindex, row))
+
+        nonempty_index = nonempty_columns[0][0]
```
thisisdhaas, line 116: is `nonempty_index` the same as `depth` in your export logic?

orchestra/todos/import_export.py
```diff
@@ -0,0 +1,122 @@
...
+        elif nonempty_index <= len(parent_items):
+            parent_items = parent_items[:nonempty_index]
+            parent_items[-1].insert(0, item)
+        else:
+            raise SpreadsheetImportError(
+                'Row {} is not a child of a previous row: {}'.format(
+                    rowindex, row))
+        parent_items.append(item['items'])
+
+    todo_list_template.todos = todos
```
thisisdhaas, line 144: Overall I find this function pretty tough to read, and in particular it doesn't seem to neatly invert the 'depth' logic you have on export. I wonder if it would be more readable to do this recursively,  like "parse rows of the same depth into a list of items, and recursively build an item's children whenever you encounter a new depth". Maybe you have to iterate over the spreadsheet in reverse to do that?

Some specific questions on the code:
* this line 116 sets `todo_list_template.todos = todos`, but `todos` is just set to the last row in the spreadsheet at this point, right? Is that related to my question above (i.e., does the last row in the spreadsheet correspond to a single root item node for all templates?)
* Lines 107-109 are pretty hard to interpret. At the very least let's add a comment explaining them.
* On line 108, are we losing the parent_items to the right of `nonempty_index`? Or is this the logic that figures out that those items aren't actually parent items, and they're already included in `item`?

orchestra/todos/urls.py
```diff
@@ -29,4 +29,7 @@
     url(r'^worker_task_recent_todo_qas/$',
         views.worker_task_recent_todo_qas,
         name='worker_task_recent_todo_qas'),
+    url(r'^import_todo_list_template_from_spreadsheet/(?P<pk>[0-9]+)/$',
```
thisisdhaas: feels a bit more RESTful to structure this as something like `todolist_templates/<pk>/import`, no?

orchestra/todos/forms.py
```diff
@@ -0,0 +1,6 @@
+from django import forms
+
+
+class ImportTodoListTemplateFromSpreadsheetForm(forms.Form):
```
thisisdhaas, line 4: You could imagine having this form do the validation on the URL here (it should look like a google spreadsheets URL), so you don't have to do it in your `get_csv_from_spreadsheet` func.

orchestra/todos/views.py
```diff
@@ -78,6 +86,44 @@ def worker_task_recent_todo_qas(request):
...
+        if form.is_valid():
+            try:
+                todo_list_template = TodoListTemplate.objects.get(
+                    id=pk)
+                import_from_spreadsheet(
+                    todo_list_template,
+                    form.cleaned_data['spreadsheet_url'],
+                    request)
+                return redirect(
+                    'admin:orchestra_todolisttemplate_change',
```
thisisdhaas, line 249: is there a way to add an 'import successful' badge to the top of the change admin view when you redirect? I vaguely recall that being possible.

orchestra/todos/views.py
```diff
@@ -78,6 +86,44 @@ def worker_task_recent_todo_qas(request):
...
+                    pk)
+            except SpreadsheetImportError as e:
```
thisisdhaas: what about other exceptions (e.g., the `assert` statement that verifies the URL's shape)?

orchestra/tests/test_todos.py
marcua: The proper `reverse` fixed it!

orchestra/tests/test_todos.py
marcua: The `reverse` fixed it! Thanks @jumasheff!

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
marcua: That's what I get for copy-paste! Thanks for catching. Did a read of the file and cleaned it up.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
marcua: Done! 
image

I did it relatively hackily, which seems fine given:
- there's no design system for Orchestra (yet)
- this is internal admin tooling, so even less luck in importing reusable styles.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
marcua: I cleaned it up (was a bad copy/paste from an internal interface we have). I won't refactor/rename until we have a known second or third use case.

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
marcua: All better :). Thanks for catching!

orchestra/core/errors.py
marcua: Good call! Renamed.

orchestra/admin.py
marcua: That's correct! To be true to how we've used this as we beta tested it, I'd replace the second bullet with `Export the empty spreadsheet as a starting point for editing.`

It's not lost on me that creating a new template is still a little annoying. In particular, the defaults for the JSON blobs in admin aren't saveable, so you have to copy/paste something reasonable-looking from another template. I started to fix this and realized the PR was big enough and we can handle cleaner creation in a separate PR.

orchestra/google_apps/convenience.py
marcua: For transparency, this was a copy/paste job from our closed/source internal tooling for managing spreadsheets. Do you think it's reasonable to make "Reimplement/harmonize Google Sheet->CSV convenience function" a separate task/PR?

orchestra/admin.py
marcua: Bullet 3 from the [django-object-actions page](https://pypi.org/project/django-object-actions/) is relevant.
image

From a decorator perspective, the only restrictions we have on the views are staff (not superuser) access. Here's a checklist I'm using to manually verify access control
- [x] Log in as a staff member
- [x] Navigate to a template's change page (CP) in Django admin
- [x] Click `import` page and open it in a new tab (IP1) so we have a tab open with the page we get redirected to
- [x] Click `import` page and open it in a new tab (IP2) so we have a tab open with the page we get redirected to
- [x] In another tab, hijack a non-staff user
- [x] confirm you can't access Django admin
- [x] Reload the IP1 tab and get access denied on the GET
- [ ] Submit the form in the IP2 tab and get access denied on the POST (in truth I got a CSRF failure because I switched user, so technically I didn't confirm this. The same decorator protects the GET and POST `View`, FWIW.)
- [x] Go to the CP tab, click `import`, and get access denied
- [x] Go to the CP tab, click `export`, and get access denied

orchestra/models/core/models.py
marcua: That's the idea! Seems to be a nice balance of small (logical) records for which a rare recovery is slightly more expensive (compared to the physical record you suggest as an alternative).

orchestra/tests/test_todos.py
marcua: yup!

orchestra/tests/test_todos.py
marcua: Good question! Our todo list templates are dynamic, meaning the machine can skip (mark irrelevant) or remove (outright delete) entire subtrees. The conditional property function returns a dictionary of facts about a project. The `skip_if` and `remove_if` entries of any node in the `todos` tree contain conjunctive normal form (ANDs of logical operations) that operate on the facts returned from conditional property function to determine whether to skip or remove their part of the tree.

That answers what the `condition_property_function` is. We don't have to import/export it because we're only import/exporting the `todos` tree (and logical operators), not the fact-returning function.

(It goes without saying that updating the Orchestra docs on this stuff would be a huge service.)

orchestra/tests/test_todos.py
marcua: The annoying thing is that the JSON is in individual cells of the spreadsheet, so you'd have to parse the CSV, pluck out specific cells, and `json.loads()` to compare them individually. With my less elegant hack, we can compare the entire CSV output.

orchestra/tests/test_todos.py
marcua: Not that I knew this, but yes it does :).

https://docs.djangoproject.com/en/3.1/topics/testing/tools/#the-test-client

image

orchestra/todos/forms.py
marcua: I've cleaned up/removed the assertion logic in `get_csv_from_spreadsheet` and recast it as a validation error that gets reported in the form. So not quite what you suggest, but same effect! :)

orchestra/migrations/0087_todolisttemplateimportrecord.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.13 on 2020-08-26 19:17
```
marcua: Will conflict with @jumasheff 's 0087.

orchestra/todos/urls.py
marcua: done!

orchestra/todos/views.py
marcua: ask and you shall receive

image

orchestra/todos/import_export.py
marcua: noic, did this in concert with the suggestion below.

orchestra/todos/import_export.py
marcua: noic done

orchestra/todos/import_export.py
marcua: `alternateLink` it is!

orchestra/todos/import_export.py
marcua: Yup! There's always a single root. `todo_list_template.todos` is shaped like this:
```
{
...
items: [{..., items: [...]}, {..., items: [...]}, ...]
...
}
```

orchestra/todos/import_export.py
marcua: Yup! Added a comment confirming this.

orchestra/todos/import_export.py
marcua: Thank you for this feedback. I couldn't figure out a nice way to turn my code into a recursive equivalent of the export (I have some vague ideas about using the stack instead of `parent_items`, but it feels more confusing and/or I'm being lazy/intellectually dishonest 😛). I took a stab at commenting this close to line-by-line. LMK your thoughts.

orchestra/todos/views.py
marcua: Does my new exception/recasting cleanup address this?

orchestra/migrations/0087_todolisttemplateimportrecord.py
marcua: (won't merge this until he merges his, and then I'll create a new migration)

orchestra/google_apps/convenience.py
```diff
@@ -217,6 +223,18 @@ def upload_file(parent_id, file_path, title, description, mimetype):
...
         mimetype,
+        mimetype,
         file_path
     )
     return file_metadata
+
+
+def get_google_spreadsheet_as_csv(gdoc_url, reader=csv.DictReader):
+    matches = GSPREAD_RE.match(gdoc_url)
+    if matches is None:
```
marcua, line 234: Cleaned this up at your direction --- we recast this newly minted ValueError as an ImportError in the caller, so all exceptions happen cleanly without `assert`s

orchestra/admin.py
thisisdhaas: makes sense, ty!

orchestra/google_apps/convenience.py
thisisdhaas: 🍇. Very reasonable.

orchestra/admin.py
thisisdhaas: very thorough, ty!

orchestra/models/core/models.py
thisisdhaas: follow-up: consider implementing ARIES for point-in-time recovery of spreadsheets from the logical logs.

orchestra/google_apps/convenience.py
thisisdhaas: ty! Will be good to unify this with our closed-source implementation in your follow-up task!

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
thisisdhaas: Only light pushback that there's already an `error_message` class on the div. Any idea where that's defined / if it's easy to extend? If not, the hackiness is fine.

orchestra/tests/test_todos.py
thisisdhaas: Gotcha. Another nice task for the future might be to validate on import that the `skip_if` and `remove_if` logic only references valid fields output by the `conditional_property_function` somehow.

orchestra/tests/test_todos.py
thisisdhaas: fair enough.

orchestra/tests/test_todos.py
thisisdhaas: fun!

orchestra/todos/import_export.py
thisisdhaas: v. noic(e)

orchestra/todos/views.py
thisisdhaas: seems to!

orchestra/templates/orchestra/import_todo_list_template_from_spreadsheet.html
marcua: I `git grep error_message|grep css`d (as well as a few other variants) in the repo I copy/pasted (🙈) this from and didn't find anything.

orchestra/todos/import_export.py
thisisdhaas: Okay, let's leave this code as is. I figured out how to make it recursive, but don't think rewriting it will make it much more readable, so iterative + your great comments is probably good enough.

In case you're curious, my thoughts on a recursive approach:
* The problem is that the tree of items isn't a proper datastructure, it's just a nested python dict. This means that there's no way to pop back 'up' the tree when we encounter a row of depth <= its parent.
* We could solve this by a) implementing a tree datastructure (probably overkill), or b) explicitly passing the parent links into the recursion. In pseudo-code, b roughly looks like:
   ```
   def _recursive_call(all_rows, cur_row_index, ancestor_stack):
       # all_rows = the csv parsed into a list of rows
       # cur_row_index = the row index we're looking at
       # ancestor_stack = [root_item, root_child_1, root_grandchild_x, ...], all items that are ancestors of the current row
       cur_row = all_rows[cur_row_index]
       cur_row_depth = find_first_nonempty_row(cur_row)
       cur_item = build_item_dict(cur_row)
       if not ancestor_stack: # we haven't added the root yet, do so and recurse to the next row
           _recursive_call(all_rows, cur_row_index + 1, [cur_item])
       if cur_row_depth > len(ancestor_stack):  # we have a new child, add it (and its descendants via recursion) to the parent's item list.
           ancestor_stack[-1]['items'].insert(0, _recursive_call(all_rows, cur_row_index + 1, ancestor_stack + [cur_item]))
       else: # ignoring the error cases, we are no longer the child of the latest parent, pop up the ancestor stack and try again with this row
           _recursive_call(all_rows, cur_row_index, ancestor_stack[:-1])
   ```
* Basically, recursion lets us rewrite your parent_items list to be a more intuitive object, but introduces other complexities that probably balance out the overall pain of the function.

marcua: Awesome! If nothing else, this helps us know it's possible! :)

jumasheff (approved):

PR #666, opened 2020-08-21
Todo: viewset, serializer

jumasheff:


orchestra/todos/serializers.py
```diff
@@ -89,3 +92,60 @@ class Meta:
             'creator',
             'todos')
         read_only_fields = ('id',)
+
+
+class TodoBulkCreateListSerializer(serializers.ListSerializer):
```
marcua, line 97: sanity-checking: can we replace any of the existing serializers in this file with this one? Or is that too ambitious?

orchestra/todos/serializers.py
```diff
@@ -89,3 +92,60 @@ class Meta:
...
+    def create(self, validated_data):
+        result = [self.child.create(attrs) for attrs in validated_data]
```
marcua, line 99: what's `child` in this case?

orchestra/todos/serializers.py
```diff
@@ -89,3 +92,60 @@ class Meta:
...
+        except IntegrityError as e:
+            raise ValidationError(e)
+        return result
+
+    def update(self, instances, validated_data):
+        instance_hash = {
+            index: instance for index, instance in enumerate(instances)}
+
+        result = [
+            self.child.update(instance_hash[index], attrs)
```
marcua: what does `update` do if we're calling `bulk_update` later?

orchestra/todos/serializers.py
```diff
@@ -89,3 +92,60 @@ class Meta:
...
+    def create(self, validated_data):
+        result = [self.child.create(attrs) for attrs in validated_data]
+        try:
+            self.child.Meta.model.objects.bulk_create(result)
+        except IntegrityError as e:
+            raise ValidationError(e)
+        return result
+
+    def update(self, instances, validated_data):
+        instance_hash = {
```
marcua: why do we need instance_hash?

could we instead do something like `for instance, data in zip(instances, validated_data)`?

orchestra/todos/serializers.py
```diff
@@ -89,3 +92,60 @@ class Meta:
...
+
+        try:
+            self.child.Meta.model.objects.bulk_update(result, writable_fields)
+        except IntegrityError as e:
+            raise ValidationError(e)
+
+        return result
+
+
+class BulkTodoSerializer(serializers.ModelSerializer):
```
marcua, line 124: sanity-checking: can we replace any of the existing serializers in this file with this one? Or is that too ambitious?

marcua: In particular, I'm looking at `TodoSerializer`

Since we're adding individual `create`/`update` fields below, I wonder how duplicative the functionality is.

marcua, line 124: two other things i see on TodoSerializer are `json_schema` and `qa`

can you make sure these fields are not necessary and/or migrated? additionally, can you make sure things like `fields` are consistent between the two?

orchestra/todos/views.py
```diff
@@ -178,3 +184,57 @@ class TodoListTemplateList(generics.ListCreateAPIView):
 
     serializer_class = TodoListTemplateSerializer
     queryset = TodoListTemplate.objects.all()
+
+
+def validate_ids(data, field='id', unique=True):
```
marcua: what's this validation for? is it ensuring we don't update an object more than once in a call? just making sure this is common practice and that DRF doesn't already facilitate it.

orchestra/todos/views.py
```diff
@@ -178,3 +184,57 @@ class TodoListTemplateList(generics.ListCreateAPIView):
...
+    if isinstance(data, list):
+        id_list = [int(x[field]) for x in data]
+        if unique and len(id_list) != len(set(id_list)):
+            raise ValidationError(
+                'Multiple updates to a single {} found'.format(field))
+        return id_list
+    return [data]
+
+
+class TodoListViewset(ModelViewSet):
```
marcua: similar story: can we replace other viewsets in this file, or is this functionality relatively unique?

orchestra/todos/views.py
```diff
@@ -178,3 +184,57 @@ class TodoListTemplateList(generics.ListCreateAPIView):
...
+    serializer_class = BulkTodoSerializer
+    authentication_classes = (OrchestraProjectAPIAuthentication,)
+    filter_backends = (filters.DjangoFilterBackend,)
```
marcua: what's this facilitate? is it the `filterset_fields`?

orchestra/todos/views.py
```diff
@@ -178,3 +184,57 @@ class TodoListTemplateList(generics.ListCreateAPIView):
...
+
+    def is_single_item_request_by_pk(self):
+        return (
+            self.action == 'retrieve' or
+            self.action == 'update' or
+            self.action == 'partial_update' or
+            self.action == 'destroy'
+        )
+
+    def get_queryset(self, ids=None):
```
marcua: Shows my ignorance, but I'd have thought there'd be a viewset that already does this that you can use with a listserializer. I don't actually know this is the case, but figured I'd leave a comment for you to sanity-check :)

orchestra/todos/urls.py
```diff
@@ -30,3 +31,10 @@
         views.worker_task_recent_todo_qas,
         name='worker_task_recent_todo_qas'),
 ]
+
+router = routers.SimpleRouter()
+router.register(
```
marcua: should this stuff live in the project_api rather than the todo api?

orchestra/project_api/auth.py
```diff
@@ -19,6 +19,8 @@ def has_permission(self, request, view):
         return request.user and isinstance(request.user, SignedUser)
 
 
+# TODO(murat): rename and move to common/ since it's being
```
marcua: Let's see if this is the case after we think through whether the new viewset should also live in the project api.

orchestra/tests/test_todos.py
```diff
@@ -602,3 +605,134 @@ def test_conditional_skip_remove_todos_from_template(self):
...
+
+
+class BulkTodoSerializerTests(EndpointTestCase):
+    def setUp(self):
+        super().setUp()
+        setup_models(self)
+        self.project = ProjectFactory()
+        self.step = StepFactory()
+        self.list_url = reverse('orchestra:todos:todo-api-list')
+        self.todo = TodoFactory(project=self.project)
```
marcua: i get that `step` isn't required, but I'm wondering if we can start enforcing it in the API -> it seems like all todos should have both a project and a step, at least when saving them.

orchestra/tests/test_todos.py
```diff
@@ -602,3 +605,134 @@ def test_conditional_skip_remove_todos_from_template(self):
         ]
         for todo, expected_todo in zip(todos, expected_todos):
             self._verify_todo_content(todo, expected_todo)
+
+
+class BulkTodoSerializerTests(EndpointTestCase):
```
marcua: is anything in the new viewsets/serializers not covered by this test?

orchestra/todos/views.py
marcua: My intuition is there won't be an opportunity for reuse because the authentication methods are different, so just a sanity-check. That said, if you decide this should be have like/live with other project_api viewsets, make sure the authentication methods and other settings look similar to the other project_api endpoints.

orchestra/tests/test_todos.py
marcua: worth testing that you can't access this endpoint as a logged-in user (only the project API authentication?)

orchestra/todos/views.py
jumasheff: Yep, it's for `filterset_fields`

orchestra/todos/views.py
thisisdhaas: Just creeping on this PR, but I would think that DRF should handle the "filter by pk if it's a detail view" logic out of the box. Perhaps you could do something simpler like:
```
def get_queryset(self, ids=None):
    queryset = super().get_queryset()
    if ids:
        # filter by them
    return queryset.order_by('-created_at')
```

orchestra/todos/serializers.py
jumasheff: I'll address this in the next PR.

jumasheff: The `qa` only assigns `None`. We can set `None` in the corresponding views. Again, this will be done in its own PR.

orchestra/todos/views.py
jumasheff: Removed, since it was trying to solve non–existing problem.

orchestra/todos/views.py
jumasheff: I think we can move some logic as shared functions. For example, slack notification :) In the next PR.

orchestra/todos/views.py
jumasheff: Wow! I was ignorant. Simplified.

orchestra/api_urls.py
```diff
@@ -86,3 +88,10 @@
         message_project_team,
         name='message_project_team'),
 ]
+
+router = routers.SimpleRouter()
+router.register(
+    r'todo-api', TodoListViewset, basename='todo-api'
```
marcua: should this be prefixed with `project/` so it's "part of" the project api?

orchestra/project_api/auth.py
marcua: i think we can remove this now, right?

orchestra/tests/test_todos.py
```diff
@@ -3,12 +3,14 @@
 from django.utils import timezone
```
marcua: extraneous imports

orchestra/todos/views.py
```diff
@@ -3,6 +3,7 @@
 from rest_framework import generics
 from rest_framework import permissions
 from rest_framework.response import Response
+from rest_framework.exceptions import ValidationError
```
marcua: extraneous import :)

orchestra/project_api/views.py
```diff
@@ -154,3 +160,35 @@ def message_project_team(request):
...
+    filterset_fields = ('project', 'step',)
+    queryset = Todo.objects.all()
+
+    def get_serializer(self, *args, **kwargs):
+        if isinstance(kwargs.get('data', {}), list):
+            kwargs['many'] = True
+
+        return super().get_serializer(*args, **kwargs)
+
+    def get_queryset(self):
```
marcua: you no longer need the

```
if ids:
        # filter by them
```

stuff?

orchestra/tests/test_todos.py
jumasheff: Fixed :)

orchestra/project_api/views.py
jumasheff: This was a very important heads up. Thanks! Brought it back and added a small test case.
For some reason, I had thought the `get_queriset` had been unreasonably clever.

orchestra/project_api/views.py
```diff
@@ -176,13 +176,16 @@ def get_serializer(self, *args, **kwargs):
 
         return super().get_serializer(*args, **kwargs)
 
-    def get_queryset(self):
+    def get_queryset(self, ids=None):
         queryset = super().get_queryset()
+        if ids:
+            return queryset.filter(id__in=ids).order_by('-created_at')
```
marcua:
```suggestion
            queryset = queryset.filter(id__in=ids)
```

marcua: with this change, there will be only one exit/return point from the function, so following its control logic will be easier :)

marcua (approved):

PR #681, opened 2020-09-03
Task unification crud api

jumasheff:


orchestra/orchestra_api.py
```diff
@@ -87,6 +87,36 @@ def get_project_information(project_ids):
...
+    step_slug_param = 'step__slug={}'.format(
+        step_slug) if step_slug is not None else ''
+    juncture = '&' if project_param and step_slug_param else ''
+    query_params = '?{}{}{}'.format(project_param, juncture, step_slug_param)
+
+    response = _make_api_request('get', 'todo-api', query_params)
+    return json.loads(response.text)
+
+
+def update_todos(updated_todos):
```
jumasheff, line 108: Should we implement `partial_update` instead? I bet we'll only update some fields of todos.
```
PUT: The PUT method replaces all current representations of the target resource with the request payload.

PATCH: The PATCH method is used to apply partial modifications to a resource.
```

orchestra/orchestra_api.py
```diff
@@ -87,6 +87,36 @@ def get_project_information(project_ids):
     return json.loads(response.text, object_hook=convert_key_to_int)
 
 
+def create_todos(todos):
+    response = _make_api_request('post', 'todo-api',
+                                 data=json.dumps(todos))
+    return json.loads(response.text)
+
+
+def get_todos(project_id=None, step_slug=None):
```
marcua: Is project_id optional?

orchestra/orchestra_api.py
```diff
@@ -87,6 +87,36 @@ def get_project_information(project_ids):
...
+    project_param = 'project={}'.format(
+        project_id) if project_id is not None else ''
+    step_slug_param = 'step__slug={}'.format(
+        step_slug) if step_slug is not None else ''
+    juncture = '&' if project_param and step_slug_param else ''
```
marcua:
```suggestion
    conjunction = '&' if project_param and step_slug_param else ''
```

orchestra/orchestra_api.py
marcua: Hard to say off the top of my mind. I'd base this on the engineering plan you made---do both use cases arise as part of the refactor?

orchestra/tests/test_orchestra_api.py
```diff
@@ -0,0 +1,168 @@
+import json
```
marcua: filename: test_project_api_client.py?

marcua: (i don't feel strongly)

orchestra/tests/test_orchestra_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+class TodoAPITests(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.request_client = APIClient(enforce_csrf_checks=True)
+        self.request_client.force_authenticate(user=SignedUser())
+        self.project = ProjectFactory()
+        self.step = StepFactory()
+
+    def _pause(self):
```
marcua: ?

orchestra/tests/test_orchestra_api.py
```diff
@@ -0,0 +1,168 @@
...
+            {
+                'title': 'Testing title {}'.format(x),
+                'project': self.project.id,
+                'step': self.step.id
+            } for x in range(10)
+        ]
+        result = create_todos(data)
+        self.assertEqual(len(result), 10)
+        for r in result:
+            self.assertTrue(r['title'].startswith('Testing title'))
```
marcua: should create calls return all of the data you created, or just IDs? honestly I don't know :)

orchestra/tests/test_orchestra_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Filter by project_id
+        res = get_todos(project_id=project2.id)
+        self.assertEqual(len(res), 2)
+        expected_ids = [todo3.id, todo4.id]
+        for r in res:
+            self.assertIn(r['id'], expected_ids)
+
+        # Get all todos
+        res = get_todos()
```
marcua: Now that I'm seeing it, this seems like a great way to DOS the server in production. I'm now convinced project_id is not optional :)

orchestra/project_api/tests/test_project_api.py
```diff
@@ -630,32 +610,3 @@ def test_destroy_functionality(self):
         marked_as_deleted = Todo.unsafe_objects.get(pk=self.todo.pk)
         self.assertTrue(marked_as_deleted.is_deleted)
         self.assertEqual(marked_as_deleted, self.todo)
-
-    def test_bulk_update(self):
```
marcua: In this PR, we removed update and retrieve, but not delete and create. i wonder if we should just leave these tests in this PR as-is untouched?

orchestra/tests/test_orchestra_api.py
jumasheff: I'd leave it as it is since it's a common practice in Django to return created objects.

orchestra/orchestra_api.py
jumasheff: In `project_tasks` we use only `patch`. Implemented `bulk_partial_update` :)

orchestra/orchestra_api.py
```diff
@@ -93,13 +93,14 @@ def create_todos(todos):
     return json.loads(response.text)
 
 
-def get_todos(project_id=None, step_slug=None):
+def get_todos(project_id, step_slug=None):
+    if project_id is None:
+        raise OrchestraError('project_id is required')
     project_param = 'project={}'.format(
```
marcua: can put on previous line i think :)

orchestra/tests/test_project_api_client.py
```diff
@@ -85,26 +86,17 @@ def get(url, *args, **kwargs):
...
         for r in res:
             self.assertIn(r['id'], expected_ids)
 
-        # Get all todos
-        res = get_todos()
-        self.assertEqual(len(res), 4)
-        expected_ids = [todo1.id, todo2.id, todo3.id, todo4.id]
-        for r in res:
-            self.assertIn(r['id'], expected_ids)
+        # Test project_id is required
```
marcua, line 92: (Not a comment on the test, just a reminder that we mentioned we'd verify if we require project_id on the server side as well)

orchestra/todos/views.py
```diff
@@ -143,15 +143,23 @@ def delete(self, request, *args, **kwargs):
 
     @action(detail=False, methods=['put'])
     def put(self, request, *args, **kwargs):
+        partial = kwargs.get('partial', False)
         ids = [x['id'] for x in request.data]
-        instances = self.get_queryset(ids=ids)
+        # To avoid mess, sort data and queryset by ids
```
marcua:
```suggestion
        # Sort the queryset and data by primary key so we update the correct records.
```

orchestra/todos/views.py
```diff
@@ -143,15 +143,23 @@ def delete(self, request, *args, **kwargs):
...
+        instances = self.get_queryset(ids=ids).order_by('id')
         serializer = self.get_serializer(
-            instances, data=request.data, partial=False, many=True)
+            instances, data=sorted_data, partial=partial, many=True)
         serializer.is_valid(raise_exception=True)
         self.perform_update(serializer)
         data = serializer.data
         return Response(data)
 
+    @action(detail=False, methods=['patch'])
```
marcua, line 159: do we need to do anything in a urls.py to expose this method?

jumasheff: Nope :) `action` (and `MethodMapper`) takes care of urls mapping: https://github.com/encode/django-rest-framework/blob/f744da74d2878b480220ebaf9d8117ff9b79a947/rest_framework/decorators.py#L123

orchestra/todos/views.py
```diff
@@ -190,13 +190,11 @@ def get_permissions(self):
...
             # `project_admins` group.
-            if self.request.user.groups.filter(
-                    name='project_admins').exists():
-                return BulkTodoSerializerWithQA
-            else:
-                return BulkTodoSerializerWithoutQA
-        else:
-            return super().get_serializer_class()
+            return BulkTodoSerializerWithQA
+        return BulkTodoSerializerWithoutQA
```
marcua: curiously, what happens to 
image

orchestra/todos/serializers.py
```diff
@@ -121,7 +121,12 @@ class Meta:
         list_serializer_class = TodoBulkCreateListSerializer
 
 
-class BulkTodoSerializerWithoutQA(BulkTodoSerializer):
+class ValidationOverrideMixin(object):
+    def validate(self, data):
```
marcua: what is the purpose of this override?

orchestra/tests/test_todos.py
```diff
@@ -131,6 +131,27 @@ def _verify_todo_creation(self, task, success, project, step, mock_notify):
             self.assertEqual(resp.status_code, 403)
             self.assertEqual(Todo.objects.all().count(), num_todos)
 
+    @patch('orchestra.todos.views.notify_single_todo_update')
+    def _verify_project_not_required(self, todo, mock_notify):
```
marcua: don't we want all viewsets to eventually require project ID?

marcua (approved):

PR #688, opened 2020-09-28
Speed up todo lists loading time on task page

adbharadwaj:
The PR implements three optimizations:

1. Reduce the number of SQL queries triggered by the api request

> The todo lists were taking a lot of time to load because of extra SQL queries for loading related field `step` and additional `qa` data in `Todo` model. To fix this problem, I have prefetched related field step and qa information. 
> 
> Prior to this PR, an example query was taking around ~14 seconds and triggered ~800 SQL queries.
> 
> After this fix. the same query is now taking ~1 sec and 8 SQL queries.

2. Avoid duplicate api requests to todos endpoint. Reference link: https://docs.angularjs.org/api/ng/service/$http#caching

3. Parallelize api requests on client side using `Promise.all`

### DIFF
--- a/orchestra/static/orchestra/todos/todo-list.directive.es6.js
+++ b/orchestra/static/orchestra/todos/todo-list.directive.es6.js
@@ -154,16 +154,14 @@ export default function todoList (orchestraApi) {
             }, {})
           todoList.possibleTasks = Object.values(tasks).filter(task => task.status !== 'Complete' && humanSteps.has(task.step_slug))
 
-          // TODO(marcua): parallelize requests rather than chaining `then`s.
-          todoApi.list(todoList.projectId).then((todos) => {
-            todoListTemplateApi.list().then((templates) => {
-              todoQaApi.workerTaskRecentTodoQas(todoList.taskId).then((todoQas) => {
-                todoList.todoQas = todoQas
-                todoList.templates = templates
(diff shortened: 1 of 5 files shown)


marcua (approved):

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -154,16 +154,14 @@ export default function todoList (orchestraApi) {
...
-                todoList.todoQas = todoQas
-                todoList.templates = templates
-                todoList.todos = todoList.transformToTree(todos)
-                todoList.ready = true
-              })
-            })
+          const p1 = todoListTemplateApi.list()
+          const p2 = todoQaApi.workerTaskRecentTodoQas(todoList.taskId)
+          const p3 = todoApi.list(todoList.projectId)
+          Promise.all([p1, p2, p3]).then((values) => {
```
marcua: curiously can you destructure the array like so?
```suggestion
          Promise.all([p1, p2, p3]).then(([templates, todoQas, todos])) => {
```

adbharadwaj: Nice suggestion :) I have restructured it in the next commit.

marcua (approved):

PR #694, opened 2020-10-19
Fix Task page freezing due to slow todolist rendering

adbharadwaj:
The change makes sure that the children todo items are rendered only if (`ng-if`) the parent todo item is expanded.

Before change:
Screen Shot 2020-10-19 at 7 37 01 AM

After change:
Screen Shot 2020-10-19 at 7 32 15 AM


orchestra/static/dist/main.js
```diff
@@ -66963,7 +66963,7 @@ function todoQaList() {
 /* 235 */
```
marcua, line 66963: QA checklist
- [x] checking on/off
- [x] marking relevant/irrelevant, removing
- [x] approving/disapproving
- [x] creating the Slack summary

marcua (approved):

PR #699, opened 2020-11-02
orchestra_api fixes

jumasheff:
Changes introduced:

- Fixes `create_todos` by adding {'Content-type':'application/json'}
- Adds `additional_data` JSON field serialization
- Enhances `get_todos` by making it possible to pass kwargs for generic filtering
- Adds QueryParamsFilterBackend to GenericTodoViewset for enabling generic filtering
- Fixes `update_todos` by adding {'Content-type':'application/json'}
- Fixes BulkTodoSerializer's update method and its validator


orchestra/migrations/0091_auto_20201023_0912.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.13 on 2020-10-23 09:12
```
jumasheff: Rename the migration file

jumasheff: Make sure that the problem has already been solved. We already have other places where we serialize/deserialize json-fields

jumasheff: Couldn't find usage examples where we build querysets using nested fields of a JSONField, e.g. `queryset.filter(some_json_field__nested_field='value')`

orchestra/todos/views.py
```diff
@@ -134,7 +135,7 @@ class GenericTodoViewset(ModelViewSet):
     classes can select their own.
     """
     serializer_class = BulkTodoSerializer
-    filter_backends = (filters.DjangoFilterBackend,)
+    filter_backends = (QueryParamsFilterBackend, filters.DjangoFilterBackend)
```
marcua: if you have QueryParamsFilterBackend, do you need DjangoFilterBackend or filterset_fields?

orchestra/todos/filters.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework import filters
+
+
+class QueryParamsFilterBackend(filters.BaseFilterBackend):
```
marcua, line 6: I think we're OK, but does this open us up to any injection attacks? If it does, what kind of sanitization should we be doing?

orchestra/orchestra_api.py
```diff
@@ -89,24 +89,40 @@ def get_project_information(project_ids):
 
 def create_todos(todos):
     response = _make_api_request('post', 'todo-api',
+                                 headers={'Content-type': 'application/json'},
                                  data=json.dumps(todos))
     return json.loads(response.text)
 
 
-def get_todos(project_id, step_slug=None):
+def _convert_filters_to_query_params(filters_dict):
```
adbharadwaj: May be loop over `additional_data` keys to make sure we dont convert filters that are not part of the additional data?
This might solve the problem of injection that @marcua talked about.

orchestra/todos/serializers.py
```diff
@@ -85,14 +85,17 @@ class BulkTodoSerializer(serializers.ModelSerializer):
...
+    additional_data = serializers.JSONField(required=False)
 
     # TODO(murat): Remove this validation when project
     # becomes a required field in models
     def validate(self, data):
-        if data.get('project') is None:
-            raise serializers.ValidationError(
-                {'project': ['project should be supplied.']}
-            )
+        step = data.get('step')
```
adbharadwaj, line 93: What if both project and step are None? Is there a case when this will happen?

orchestra/todos/serializers.py
```diff
@@ -110,7 +113,8 @@ def create(self, validated_data):
         return instance
 
     def update(self, instance, validated_data):
-        validated_data = self._set_step_to_validated_data(validated_data)
+        if validated_data.get('step') is not None:
```
adbharadwaj, line 116: Just curious: when will step be None here?

adbharadwaj (commented):
Very close! Left 1 comment and 2 questions related to "Fixes `BulkTodoSerializer`'s update method and its validator". Overall, it would be helpful if you could describe what was fixed in `BulkTodoSerializer`.

orchestra/todos/filters.py
jumasheff:
```
Django’s querysets are protected from SQL injection since their queries are constructed using query parameterization. A query’s SQL code is defined separately from the query’s parameters. Since parameters may be user-provided and therefore unsafe, they are escaped by the underlying database driver.
Django also gives developers power to write raw queries or execute custom sql. These capabilities should be used sparingly and you should always be careful to properly escape any parameters that the user can control. In addition, you should exercise caution when using extra() and RawSQL.
```
https://docs.djangoproject.com/en/3.1/topics/security/#sql-injection-protection

According to the docs, `extra()` and `RawSQL` methods/functions are unsafe, but querysets are injection-proof.
In our case, we are not using `extra()` and `RawSQL`.

Besides, a given serializer only deals with the fields specified 1) in a serializer's meta class 2) `filterset_fields` of a view, other fields are ignored (as well as nested fields, like 'todo__project__id', unless they are explicitly specified in the serializer's fields or in filterset_fields).

Here's a good article from Jacob Kaplan-Moss on sql-injection prevention: https://jacobian.org/2020/may/15/preventing-sqli/ (by the way, it contains a good idea for our upcoming hackathon).

orchestra/orchestra_api.py
jumasheff: Currently it ignores filters like {'additional_data__some_nested_field': 'value'}` 😛

orchestra/todos/serializers.py
jumasheff: When we are updating a todo partially, the update payload may be lacking both, `step` and `project`, and can only contain, say, `id` and `order`.

orchestra/todos/serializers.py
jumasheff: When we are partially updating todos.

orchestra/todos/views.py
jumasheff: We no longer need `DjangoFilterBackend`. Removed.

jumasheff:
> Very close! Left 1 comment and 2 questions related to "Fixes `BulkTodoSerializer`'s update method and its validator". Overall, it would be helpful if you could describe what was fixed in `BulkTodoSerializer`.

2 things:
1) Added `additional_data = serializers.JSONField(required=False)` so that it's dealt as a JSON field, not as a string.
2) Made it possible to partially update todos without specifying `project__id` and `step__slug`

orchestra/orchestra_api.py
adbharadwaj: May be I am missing something. How is it ignoring filters like `{'additional_data__some_nested_field': 'value'}`?

orchestra/tests/test_query_params_filter.py
```diff
@@ -0,0 +1,93 @@
...
+        params = request.query_params.dict()
+        qs_kwargs = backend._get_filter_kwargs(view, params)
+        filterset_fields_kwargs = backend._get_filterset_fields_kwargs(
+            view, params, qs_kwargs)
+        return qs_kwargs, filterset_fields_kwargs
+
+    def test_random_key_values(self):
+        url_params = build_url_params(
+            123, 'slug', **{'some_random_key': 'some_value'})
+        qs_kwargs, _ = self._get_qs_kwargs(DummyView, url_params)
```
adbharadwaj: Is there a way to test this on `GenericTodoViewset` instead on a `DummyView`? It would be good if these tests fail if a changes allows non-serializer fields to be passed?

orchestra/todos/views.py
```diff
@@ -134,8 +134,8 @@ class GenericTodoViewset(ModelViewSet):
     classes can select their own.
     """
     serializer_class = BulkTodoSerializer
-    filter_backends = (filters.DjangoFilterBackend,)
-    filterset_fields = ('project', 'step__slug',)
+    filter_backends = (QueryParamsFilterBackend,)
```
marcua, line 137: how does this interact with the comment above?

```
Note: this doesn't support nested__fields. To support them,
    add `filterset_fields` iterable to the view.
```

orchestra/todos/filters.py
```diff
@@ -0,0 +1,46 @@
+from ast import literal_eval
+
+from rest_framework import filters
+
+
+class QueryParamsFilterBackend(filters.BaseFilterBackend):
+    """
+    Takes queryparams from a URL, serializes into a python data structure
+    and passes it as a queryset arguments.
+    Note: this doesn't support nested__fields. To support them,
```
marcua, line 14: I know one of the intentions is to be able to filter on `additional_data`'s fields. Based on this comment, don't we have to be able to support nested filters? If so, how can we do that given the `filterset_fields` iterable is set in the view?

jumasheff: Improved comments :)
We'll support generic JSONField lookups after migrating to Django 3.1, yay!
https://docs.djangoproject.com/en/3.1/releases/3.1/#jsonfield-for-all-supported-database-backends

orchestra/orchestra_api.py
jumasheff: Consider using this:
```
import urllib.parse
dict = {
    "title": "Hello world",
    "anonymous": False,
    "needautocategory": True
}
urllib.parse.urlencode(dict)
```

orchestra/orchestra_api.py
```diff
@@ -89,24 +89,39 @@ def get_project_information(project_ids):
...
-    if project_id is None:
-        raise OrchestraError('project_id is required')
-    project_param = 'project={}'.format(project_id)
+def build_url_params(project_id, step_slug, **filters):
+    project_param = 'project__id={}'.format(project_id) if project_id else ''
     step_slug_param = '&step__slug={}'.format(
         step_slug) if step_slug is not None else ''
-    query_params = '?{}{}'.format(project_param, step_slug_param)
-
+    additional_filters = '&q={}'.format(json.dumps(filters))
```
marcua: minor, and feel free to push back: you could do this only if filters is nonempty :)

orchestra/todos/filters.py
```diff
@@ -0,0 +1,58 @@
...
+                if field in params and field not in qs_kwargs:
+                    filterset_kwargs[field] = params[field]
+        return filterset_kwargs
+
+    def _get_params(self, request, view):
+        """
+        Here we get JSON encoded params and encode them into Python objects.
+        """
+        params = request.query_params.dict()
+        json_params = params.pop(self._get_query_params_prefix(view), '{}')
```
marcua, line 47: is it considered good practice to modify the query params? as in, should this `pop` be a `get`?

orchestra/project_api/tests/test_project_api.py
```diff
@@ -575,12 +587,22 @@ def test_get_list_of_todos_with_filters(self):
...
         self.assertEqual(resp.json()[0]['step'], self.todo_with_step.step.slug)
 
-        url_with_filters = '{}?project={}&step__slug={}'.format(
+        url_with_filters = '{}?project__id={}&step__slug={}'.format(
             self.list_url, self.project.id, self.todo_with_step.step.slug)
         resp = self.request_client.get(url_with_filters)
         self.assertEqual(resp.status_code, 200)
         self.assertEqual(resp.json()[0]['step'], self.todo_with_step.step.slug)
 
+        ids_to_filter_by = [self.todo.id, self.todo_with_step.id]
```
adbharadwaj, line 600: What if you pass an id that does not exist? Having that test might be useful! :)

orchestra/tests/test_query_params_filter.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        title = 'Some title'
+        project_id = 123
+        url_params = build_url_params(
+            project_id,
+            None,
+            **{
+                'title': title,
+                'nonexistent_field': True})
+        qs_kwargs, filterset_fields_kwargs = self._get_qs_kwargs(
```
adbharadwaj: I know that `_get_qs_kwargs ` uses the same logic as `filter_queryset` in `QueryParamsFilterBackend `. But I feel like if for some reason we change the logic in `filter_queryset`, the tests would not catch it. How about we create `_get_qs_kwargs`  in `QueryParamsFilterBackend`, use it in `filter_queryset` and test it directly. What do you think?

orchestra/project_api/tests/test_project_api.py
```diff
@@ -562,7 +574,7 @@ def test_get_single_todo_by_pk(self):
         self.assertEqual(resp.status_code, 200)
 
     def test_get_list_of_todos_with_filters(self):
-        url_with_project_filter = '{}?project={}'.format(
+        url_with_project_filter = '{}?project__id={}'.format(
```
adbharadwaj, line 577: I know that we might end up with long method names but can we create separate tests for different scenarios e.g. when project_id is provided with/without id__in? I feel tests will be more readable that way. 

If you feel otherwise, can you add comments about each scenario right before you test it.

orchestra/todos/filters.py
jumasheff: Addressed :) Now we mutate a copy of the object.

orchestra/project_api/tests/test_project_api.py
jumasheff: Done :)

orchestra/tests/test_query_params_filter.py
jumasheff: I like your suggestion. Fixed :)

orchestra/project_api/tests/test_project_api.py
jumasheff: Done :)

orchestra/todos/filters.py
marcua: Curiously, why do we want to incur a copy when we can use `get` and leave the original unpurturbed?

jumasheff: I just realized, that the `pop` version was better :)
Suppose, we passed `?project__id=8485&q={"title": "some title"}`
So in params we will have `{'project__id': '8485', 'q': '{"title": "some title"}'}`
When we **get** `q`, we leave the original 'q' key with its unconverted value and we end up with `{'project__id': '8485', 'q': '{"title": "some title"}', 'title': 'some title'} after converting things into dict and merging back.

jumasheff: So, my suggestion is we
a) clone `request.query_params.copy().dict()` pop `q`, convert and merge the converted dict into params
b) clone `request.query_params.dict().copy()` pop `q`, convert and merge the converted dict into params
c) we don't touch `request.query_params.dict()`, and create a third dictionary where we merge `params` and `converted`. But we'll end up with `q` there:
```
        params = request.query_params.dict()
        json_params = params.get(self._get_query_params_prefix(view), '{}')
        converted = json.loads(json_params)
        result = params.copy()
        result.update(converted)
        return result
```

adbharadwaj (approved):
a!

PR #700, opened 2020-11-06
Add get_todos_by_ids

jumasheff:


adbharadwaj (changes requested):
Can we add a test for the new api? Or are you planning to add it later?

adbharadwaj (approved):
a!

PR #714, opened 2020-11-20
Updates hijack banner styles to be cleaner and collapsible.

thisisdhaas:
Blast from the past! This PR applies nicer hijack styles to our hijack banner, so it can be hidden (now to a thin but visible line) without refreshing the page, and re-expanded easily.

Since 2020, I applied some tiny style updates to optimize for mobile. Here is the latest look:
| mobile | desktop |
|--------|---------|
| Screen Shot 2023-04-06 at 2 36 13 PM | Screen Shot 2023-04-06 at 2 35 50 PM |

### DIFF
--- a/orchestra/static/orchestra/common/css/_common.scss
+++ b/orchestra/static/orchestra/common/css/_common.scss
@@ -74,3 +74,124 @@ $main-font-stack: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helveti
      transform:rotate(360deg);
   }
 }
+
+// styles related to the hijack banner
+.hijacked-warning {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 50px;
+  padding: 5px 10px;
+  position: fixed;
+
+  &__message {
+    font-size: 14px;
+    line-height: 16px;
+    text-align: left;
+    font-weight: normal;
+
+    span {
+      font-weight: bold;
+    }
+  }
+
+  &__actions {
+    display: flex;
+    justify-content: flex-end;
+  }
+
+  &__show {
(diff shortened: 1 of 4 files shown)


orchestra/static/orchestra/common/css/_common.scss
```diff
@@ -74,3 +74,124 @@ $main-font-stack: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helveti
      transform:rotate(360deg);
```
marcua, line 74: sanity-checking we should be updating timer.css (maybe) and orchestra.css (probably)

orchestra/static/orchestra/common/css/_common.scss
```diff
@@ -74,3 +74,124 @@ $main-font-stack: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helveti
...
   }
 }
```
marcua, line 76: does the time-tracking modal and each screen from the top right menu (e.g., account settings, projman, ...) still display properly?

orchestra/static/orchestra/common/css/_common.scss
thisisdhaas: I think all are necessary to render the various pages. I updated _common.scss to make sure the css was loaded on every page, and gulp did the rest.

orchestra/static/orchestra/common/css/_common.scss
thisisdhaas: Yes! Sent you a video walkthrough offline to confirm.

marcua (approved):
yay! ty!

PR #736, opened 2021-01-20
Add orchestra api for todo templates and starting order

adbharadwaj:
Added three orchestra API:

- Get todo templates
  - Returns list of todo templates
- Create todos from a template 
  - Input is `project id`, `template slug`, and `step slug`
  - Returns a dictionary with keys `success=True` and `todos=List of newly created todos`
- Add step id field to the list of steps returned with get project information api


orchestra/project_api/views.py
```diff
@@ -157,10 +164,93 @@ def message_project_team(request):
...
+                'error': str(e)
+            }
+        }
+
+
+@api_endpoint(methods=['POST'],
+              permissions=(IsSignedUser,),
+              logger=logger,
+              auths=(OrchestraProjectAPIAuthentication,))
+def todo_sections_starting_order(request):
```
marcua: @jumasheff is doing the code review, so i'll just focus on this one endpoint to make sure I understand its purpose.

I can't think of an experience in Orchestra or the B12 product that requires this order from Orchestra, as the product currently decides on the fixed section order on its end. Since Orchestra doesn't model sections deeply (maybe it should---right now they are just strings), it seems like the metadata for the section order doesn't have a great home in Orchestra without more data modelling.

Because of the lack of model, the implementation here orders sections by count, which is different from the behavior in the product. I'm open to ideas, but I can imagine two worlds
- Make it so sections are just strings in Orchestra. The caller (e.g., B12 product) can determine their display order.
- Push section metadata like section order into Orchestra's data model for to-do lists, and as a result, add that information to the existing to-do endpoints.

orchestra/tests/test_todos.py
```diff
@@ -581,6 +587,7 @@ def test_update_todos_from_todolist_template_success(self):
                        project=self.project.id,
                        step=self.step.slug),
         ]
+        print(expected_todos)
```
jumasheff: Remove `print`

orchestra/project_api/views.py
```diff
@@ -157,10 +164,93 @@ def message_project_team(request):
...
+    except Exception as e:
+        return {
+            'success': False,
+            'errors': {
+                'error': str(e)
+            }
+        }
+
+
+@api_endpoint(methods=['POST'],
```
jumasheff: Why `POST`, and not `GET`?

orchestra/tests/test_todos.py
adbharadwaj: Removed :)

orchestra/project_api/views.py
adbharadwaj: Got rid of the API itself.

orchestra/project_api/views.py
adbharadwaj: Got rid of the API.

orchestra/project_api/views.py
```diff
@@ -157,10 +163,66 @@ def message_project_team(request):
...
     permission_classes = (IsSignedUser,)
     authentication_classes = (OrchestraProjectAPIAuthentication,)
+
+
+class TodoTemplatesList(generics.ListAPIView):
+    permission_classes = (IsSignedUser,)
+    authentication_classes = (OrchestraProjectAPIAuthentication,)
+
+    serializer_class = TodoListTemplateSerializer
+    queryset = TodoListTemplate.objects.all()
```
jumasheff, line 228: `queryset = TodoListTemplate.objects.filter(is_deleted=False)`

adbharadwaj: I don't think we will need this because we are inheriting from BaseModel which inherits from BaseModelManager.

https://github.com/b12io/orchestra/blob/e2c993080cc0acda87269bad3e9d53272b20efa7/orchestra/models/core/models.py#L597

https://github.com/b12io/orchestra/blob/e2c993080cc0acda87269bad3e9d53272b20efa7/orchestra/utils/models.py#L53-L64

https://github.com/b12io/orchestra/blob/e2c993080cc0acda87269bad3e9d53272b20efa7/orchestra/utils/models.py#L42-L50

BaseModelManager sets the default query to `is_deleted=False`

adbharadwaj: Just to be sure, I confirmed by deleting a todolist template and made sure that we don't need `is_deleted=False`

jumasheff: Wow! Ignore my suggestion. A!

jumasheff (approved):

PR #737, opened 2021-01-21
Pin py-moneyed dependency

adbharadwaj:
Pin py-moneyed dependency to latest version (1.1.0)


marcua (approved):
lgtm, assuming you've tested the following
* before migration, visually inspect PayRates in Django admin and confirm they are in $
* run migration
* confirm that all PayRates are still in $
* change one to a different currency ensure it saves OK!

PR #744, opened 2021-02-12
Migrating Todo completed and skipped_datetime to status field

adbharadwaj:
This PR does two things:
1. Migrates Todo model to use `status` field instead of `completed` and `skipped_datetime` field
2. Fixes a bug where we don't load the updated todolist properly after adding a new todolist template.

QA:

- [x] Add a todo
- [x] Add a todo from template
- [x] Remove a todo and todolist
- [x] Mark as completed/pending -- both flat and nested todo (reload and make sure changes persist)
- [x] Mark as relevant/not relevant -- both flat and nested todo (reload and make sure changes persist)
- [x] In the `QA` box, approve of one checklist item and disapprove of another. Add a comment to the disapproved item.
- [x] Reload the task. The approved/disapproved checklist items should persist, as should the comment.
- [x] Click `Copy QA summary` and paste it in a text editor. The list should reflect your approval/disapproval/comment.


orchestra/management/commands/migrate_todos_status.py
```diff
@@ -0,0 +1,76 @@
+from django.core.management.base import BaseCommand
```
marcua, line 1: I assume @jumasheff is reviewing, so I'll just skim.

One q: on approx ~1M to-dos, how quickly does this script run?

orchestra/management/commands/migrate_todos_status.py
```diff
@@ -0,0 +1,76 @@
...
+            except IndexError:
+                end_pk = queryset.values_list('pk', flat=True).last()
+
+            yield queryset.filter(pk__gt=start_pk).filter(pk__lte=end_pk)
+
+            start_pk = end_pk
+
+    def bulk_update_todo_status(self, queryset, status, batch_size):
+        num_rows = queryset.count()
+        logger.info('Started updating {} todos with status = {}'.format(
```
marcua: throughout: use `%`s in logger strings rather than `format` so that messages get grouped appropriately :)

orchestra/management/commands/migrate_todos_status.py
```diff
@@ -0,0 +1,76 @@
...
+            num_rows, status))
+        for chunk in self.chunked_queryset(queryset, batch_size):
+            chunk.only('id', 'status').update(status=status)
```
marcua: as you've implemented this, because you have a queryet and call `update` on it with a constant value, you've pushed all the work into the database. this is great---there's basically only three queries you run, and the DB is pretty effective at making the update to each of the relevant columns. i'd remove the chunking/batch size entirely! :)

you'd normally have batching/chunking in cases where you need to bring the queryset into memory and update individual rows after running some business logic. in this case, you pushed the entire logic into the `filter`, which is awesome!

orchestra/models/core/models.py
```diff
@@ -706,9 +706,11 @@ class Status(ChoicesEnum):
     details = models.TextField(null=True, blank=True)
     section = models.CharField(max_length=255, null=True, blank=True)
     order = models.IntegerField(null=True, blank=True)
+    # DEPRECATED FIELD. [Use status instead.]
```
marcua: The []'s in the docs were to indicate optional :)
```suggestion
    # DEPRECATED FIELD. Use status instead.
```

orchestra/models/core/models.py
```diff
@@ -706,9 +706,11 @@ class Status(ChoicesEnum):
...
     completed = models.BooleanField(default=False)
     start_by_datetime = models.DateTimeField(null=True, blank=True)
     due_datetime = models.DateTimeField(null=True, blank=True)
+    # DEPRECATED FIELD. [Use status instead.]
```
marcua:
```suggestion
    # DEPRECATED FIELD. Use status instead.
```

orchestra/static/orchestra/todos/constants.es6.js
```diff
@@ -0,0 +1,3 @@
+export const PENDING_STATUS = 0
```
marcua, line 1: i forget---do we not have a way to pass enum values to the frontend in open source Orchestra?

orchestra/todos/api.py
```diff
@@ -83,14 +82,13 @@ def _add_template_todo(
         additional_data):
     remove = _to_exclude(conditional_props, template_todo.get('remove_if', []))
     if not remove:
-        if parent_todo.skipped_datetime:
-            skipped_datetime = parent_todo.skipped_datetime
+        if parent_todo.status == Todo.Status.DECLINED.value:
+            to_skip = True
```
marcua, line 86: lol was to_skip previously not getting used anywhere?

orchestra/todos/views.py
```diff
@@ -47,8 +47,7 @@ def update_todos_from_todolist_template(request):
     try:
         add_todolist_template(todolist_template_slug, project_id, step_slug)
         todos = Todo.objects.filter(
-            project__id=project_id,
-            step__slug=step_slug).order_by('-created_at')
+            project__id=project_id).order_by('-created_at')
```
marcua, line 50: why this change?

orchestra/utils/common_helpers.py
```diff
@@ -33,11 +34,17 @@ def get_relevance_and_completion_changes(old_todo, new_todo):
...
     # to avoid triggering any slack messages
     todo_change = ''
-    if old_todo.completed != new_todo.completed:
-        todo_change = 'complete' if new_todo.completed else 'incomplete'
-    elif old_todo.skipped_datetime != new_todo.skipped_datetime:
-        todo_change = 'not relevant' \
-            if new_todo.skipped_datetime else 'relevant'
+    if new_todo.status != old_todo.status:
+        if new_todo.status == Todo.Status.COMPLETED.value and \
+                old_todo.status == Todo.Status.PENDING.value:
```
marcua, line 39: the old logic doesn't look at the old value here, in case it wasn't pending before. i think we should preserve that.

orchestra/utils/common_helpers.py
```diff
@@ -33,11 +34,17 @@ def get_relevance_and_completion_changes(old_todo, new_todo):
...
+            todo_change = 'complete'
+        if new_todo.status == Todo.Status.PENDING.value and \
+                old_todo.status == Todo.Status.COMPLETED.value:
```
marcua, line 42: same

orchestra/utils/common_helpers.py
```diff
@@ -33,11 +34,17 @@ def get_relevance_and_completion_changes(old_todo, new_todo):
     # When activity_log is updated, `todo_change = ''`
     # to avoid triggering any slack messages
     todo_change = ''
-    if old_todo.completed != new_todo.completed:
-        todo_change = 'complete' if new_todo.completed else 'incomplete'
-    elif old_todo.skipped_datetime != new_todo.skipped_datetime:
-        todo_change = 'not relevant' \
-            if new_todo.skipped_datetime else 'relevant'
+    if new_todo.status != old_todo.status:
+        if new_todo.status == Todo.Status.COMPLETED.value and \
```
marcua, line 38: should any of the stuff below be `elif` to be more explicit?

orchestra/management/commands/migrate_todos_status.py
jumasheff: Example: `logger.info('Started updating %s todos with status = %s', num_rows, status)`.

orchestra/todos/api.py
jumasheff: I think it was used here: `skipped_datetime = timezone.now() if to_skip else None` (line 91)

orchestra/management/commands/migrate_todos_status.py
adbharadwaj: 2-3 minutes.

orchestra/static/orchestra/todos/constants.es6.js
adbharadwaj: I couldnt find an example. Maybe @paopow or @jumasheff know a way?

orchestra/todos/api.py
adbharadwaj: It was being used as @jumasheff pointed out.

orchestra/todos/views.py
adbharadwaj: There was a bug in code where we were showing only the todos associated with the step role for which the todolist template was recently added (will add to PR description). I can show it to you live.

orchestra/utils/common_helpers.py
adbharadwaj: The old logic did not need to look for old value because we had already checked if new and old value are not same (`old_todo.completed != new_todo.completed`).

Given that completed (True/False) and skipped_datetime (Null/Not Null) field could only have two states, old value could be implicitly inferred from the new value. This is not the case with status field where we can have three types of values (pending/declined/completed).

orchestra/utils/common_helpers.py
adbharadwaj: same ^^ :)

orchestra/management/commands/migrate_todos_status.py
adbharadwaj: Fixed

orchestra/utils/common_helpers.py
adbharadwaj: Fixed

orchestra/todos/api.py
marcua: ah perfect! silly me! :)

marcua (approved):
lgtm! thanks for the iterations!

PR #762, opened 2021-03-26
Excludes completed tasks/projects from autostaffing.

thisisdhaas:
We experienced an issue with autostaffing where a StaffbotRequest for a task in a completed project was in the DONE_SENDING_INQUIRIES state rather than the COMPLETED state. This led autostaff to repeatedly try (and fail) to staff the task.

This fix excludes RequestInquiries for completed tasks / projects from staffbot's consideration. It refactors logic we've previously used to filter those requests out of the Available Tasks table.

### DIFF
--- a/orchestra/communication/staffing.py
+++ b/orchestra/communication/staffing.py
@@ -127,12 +127,21 @@ def check_responses_complete(request):
              .format(request.task)))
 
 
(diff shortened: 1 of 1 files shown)


marcua (approved):
lgtm assuming tests pass

PR #765, opened 2021-03-31
Fix slack messages for todo updates and addition

adbharadwaj:
This PR fixes the slack notifications when todos are updated or added. 

Bug explanation:

- [create_todos API](https://github.com/b12io/orchestra/blob/main/orchestra/orchestra_api.py#L90) allows users to pass a list of new todos to create using [TodoBulkCreateListSerializer](https://github.com/b12io/orchestra/blob/main/orchestra/todos/serializers.py). This means that serializer data passed to [perform_create](https://github.com/b12io/orchestra/pull/765/files#diff-b3ed3eee712816634431297ca7826a7238d5888c41fdcc270f6666939de16996R194) function can be a list of Todo objects as well. The old code did not handle this scenario.
- The same issue happens with the `perform_update` function where the serializer can have a list of updated todos. Additionally, the line `old_todo = self.get_object()` returned the same data object after we saved the todo. We need to call this object before saving the serializer. Due to these issues, the update slack message was not going out.

### DIFF
--- a/orchestra/todos/views.py
+++ b/orchestra/todos/views.py
@@ -1,4 +1,5 @@
 import logging
+import copy
 
 from django.contrib import messages
 from django.contrib.admin.views.decorators import staff_member_required
@@ -12,6 +13,8 @@ from rest_framework.viewsets import ModelViewSet
 from rest_framework.response import Response
 from rest_framework.decorators import action
 from jsonview.exceptions import BadRequest
+from django.db.models.query import QuerySet
+
 
 from orchestra.core.errors import TodoListTemplateValidationError
 from orchestra.models import Task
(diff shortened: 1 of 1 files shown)


jumasheff (approved):

orchestra/todos/views.py
```diff
@@ -178,16 +180,23 @@ def patch(self, request, *args, **kwargs):
...
-        if isinstance(todo, Todo):
-            old_todo = self.get_object()
-            notify_single_todo_update(
-                self.request.user, old_todo, todo)
+        old_data = copy.deepcopy(serializer.instance)
+        old_todos = list(old_data) if isinstance(
+            old_data, QuerySet) else [old_data]
+        data = serializer.save()
+        todos = data if isinstance(data, list) else [data]
+        for idx, todo in enumerate(todos):
```
marcua: two thoughts:
1. instead of enumerating `todos` and indexing `old_todos` could you loop through `old_todos` instead?
2. if 1 doesn't work and you need to test the todos in `todos`, you can avoid the enumerate/index by doing `for old_todo, new_todo in zip(old_todos, todos)` which feels a bit cleaner.

adbharadwaj: Avoided the enumerate/index by doing `for old_todo, new_todo in zip(old_todos, todos)`.

marcua (approved):

PR #767, opened 2021-04-19
Rename team messages to scratchpad

adbharadwaj:
The name “Team Messages” misleads the experts to start communicating through Team messages google document. Instead, the best way to communicate is To-dos and Slack (if required) as much as possible. As a solution, we want to rename the document to Scratchpad.


orchestra/static/orchestra/common/components/team-messages/team-messages.html
```diff
@@ -1,5 +1,5 @@
-<website-iframe title="Team Messages"
+<website-iframe title="Scratchpad"
 		id="team-messages"
```
marcua: if you care :)

docs/source/api.rst
```diff
@@ -54,7 +54,7 @@ All requests must be signed using `HTTP signatures <http://tools.ietf.org/html/d
               "id": 123,
```
marcua, line 54: if you `git grep -i team|grep messages` in the root of this repository, does anything relevant pop up?

adbharadwaj: Screen Shot 2021-04-19 at 4 45 58 PM

adbharadwaj: aah that was so much better grep command than what i was running :P

orchestra/static/orchestra/common/components/team-messages/team-messages.html
adbharadwaj: care++!!

docs/source/api.rst
marcua: ooo perfect! and i just caught a small bug in my grep: `git grep -i team|grep -i messages` (added a second -i). hopefully it doesn't uncover anything!

orchestra/migrations/0094_auto_20210419_1747.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.13 on 2021-04-19 17:47
```
paopow: rename the file to `0094_rename_team_messages_url.py` or sth along the same line?

docs/source/api.rst
adbharadwaj: The new grep was showing changes from these files:
```
docs/source/bots.rst
orchestra/frontend/build/static/js/2.564c31ea.chunk.js.map
```

I checked and found nothing relevant.

paopow (approved):

PR #770, opened 2021-04-22
Run google drive folder creator in another thread

jumasheff:
This way we decreased create_project_with_tasks' execution time from 9.6 seconds to 2.2 seconds (in dev environment).

### DIFF
--- a/orchestra/project.py
+++ b/orchestra/project.py
@@ -1,8 +1,15 @@
+import threading
 from orchestra.communication.slack import create_project_slack_group
 from orchestra.google_apps.convenience import create_project_google_folder
 from orchestra.models import Project
 from orchestra.models import WorkflowVersion
 from orchestra.utils.task_lifecycle import create_subsequent_tasks
+
+
+def _run_create_project_google_folder_in_background(project):
+    t = threading.Thread(target=create_project_google_folder, args=(project,))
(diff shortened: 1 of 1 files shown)


orchestra/project.py
```diff
@@ -1,10 +1,17 @@
+import threading
```
marcua, line 1: Curiously, are there any asynchrony concerns? For example, if a machine task runs immediately after project creation (as we have in our internal workflows) and needs the folder, will it try to access the folder before it's created?

orchestra/project.py
```diff
@@ -1,10 +1,17 @@
...
 from orchestra.communication.slack import create_project_slack_group
 from orchestra.google_apps.convenience import create_project_google_folder
 from orchestra.models import Project
 from orchestra.models import WorkflowVersion
 from orchestra.utils.task_lifecycle import create_subsequent_tasks
 
 
+def _run_create_project_google_folder_in_background(project):
+    t = threading.Thread(target=create_project_google_folder, args=(project,))
+    t.daemon = True
```
marcua: curiously, what's this for?

jumasheff: I misunderstood the docs. Now I see that `daemon = True` is not what we want. Removed.

PR #787, opened 2021-06-30
Load autostaff worker hours once per run

marcua:
* Bug
  * Our autostaffing logic loops over all staffable requests and checks if an expert has available hours for the day. It queries the database for availability on each request.
  * As a result, if an expert increases her availability while autostaffing is running across all staffable tasks, or if autostaffing is running across a midnight barrier, a staffing request that is lower-priority might be the one that is checked immediately after new availability is discovered. This causes a lower-priority request to be staffed.
* Solution: Compute worker hours availability before autostaffing begins and use that consistent value in every staffing decision.
* Test that it worked
  * Run autostaffing and find that it doesn't staff anyone
  * Run autostaffing and concurrently increase an expert's availability
  * Ensure no tasks were staffed
  * Run autostaffing again, and ensure that the highest-priority tasks are staffed

### DIFF
--- a/orchestra/communication/staffing.py
+++ b/orchestra/communication/staffing.py
@@ -141,6 +141,24 @@ def _exclude_inactive_staffbot_requests(request_queryset):
             .exclude(task__status=Task.Status.ABORTED))
 
 
+def _get_worker_allowed_hours():
+    today = timezone.now().date()
+    today_abbreviation = ['mon', 'tues', 'wed', 'thurs', 'fri', 'sat', 'sun'][
+        today.weekday()]
+    return {
+        availability.worker.id:
+        min(
+            getattr(
+                availability,
+                'hours_available_{}'.format(today_abbreviation), 0),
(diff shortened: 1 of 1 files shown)


orchestra/communication/staffing.py
```diff
@@ -141,6 +141,24 @@ def _exclude_inactive_staffbot_requests(request_queryset):
             .exclude(task__status=Task.Status.ABORTED))
 
 
+def _get_worker_allowed_hours():
+    today = timezone.now().date()
+    today_abbreviation = ['mon', 'tues', 'wed', 'thurs', 'fri', 'sat', 'sun'][
+        today.weekday()]
+    return {
+        availability.worker.id:
+        min(
```
thisisdhaas, line 152: add a comment explaining the two things we're min-ning here?

thisisdhaas (approved):

orchestra/communication/staffing.py
marcua: Done! TY!

PR #798, opened 2021-07-29
Add is required function for templates

junusheva:
Add a function that we could execute to get the required field for todos created from templates


orchestra/models/core/models.py
```diff
@@ -687,6 +687,9 @@ class TodoListTemplate(TodoListTemplateMixin, BaseModel):
         conditional_property_function (str)
             A JSON blob containing the path to and name of a python method
             that will return the preconditions to prune the created todos.
+        is_todo_required_function (str)
+            A JSON blog containing the path to and name of a python method
```
paopow, line 691: A JSON blog -> A JSON blob

orchestra/models/core/models.py
```diff
@@ -687,6 +687,9 @@ class TodoListTemplate(TodoListTemplateMixin, BaseModel):
         conditional_property_function (str)
             A JSON blob containing the path to and name of a python method
             that will return the preconditions to prune the created todos.
+        is_todo_required_function (str)
```
paopow, line 690: Isn't Boolean enough here? I don't think we need the same thing as conditional_property. cc @adbharadwaj in case you have other thoughts.

PR #804, opened 2021-08-13
Speed up project management page

adbharadwaj:
The project management page loads all project data even when we just need to show a selected project's info.

This PR makes two changes:

1. When a project is selected, we do not load all projects data. This speeds up the page loading.
2. When no project is selected, we continue to load all projects. So the page speed remains the same!


orchestra/static/dist/main.js
```diff
@@ -63153,12 +63150,12 @@ function projectVis($uibModal, $location, dataService, orchestraApi, crosshair,
       axisWrapper.append('span').attr('class', 'x label');
 
       scope.$on('orchestra:projectManagement:dataUpdate', vis.draw);
-      dataService.ready.then(function () {
-        if (projectId) {
-          dataService.changeProject(projectId);
-          visUtils.parentContainer.node().scrollLeft = 100;
-        }
-      });
+      if (projectId) {
```
adbharadwaj, line 63153: We shouldn't load all projects when a project is already selected!

orchestra/static/orchestra/project-management/project-management.html
```diff
@@ -8,7 +8,7 @@
         <div class="col-lg-12 col-md-12 col-sm-12">
```
marcua, line 8: should you hide the dropdown entirely instead of disabling it, which might make people think the interface is still loading?

orchestra/static/orchestra/project-management/project-vis.es6.js
```diff
@@ -51,12 +51,12 @@ export default function projectVis (
...
-        if (projectId) {
-          dataService.changeProject(projectId)
-          visUtils.parentContainer.node().scrollLeft = 100
-        }
-      })
+      if (projectId) {
+        dataService.changeProject(projectId)
+        visUtils.parentContainer.node().scrollLeft = 100
+      } else {
+        dataService.ready
```
marcua: do you have to return this or something? i don't get what this line does on its own.

orchestra/static/orchestra/project-management/project-management.html
marcua: I realize you're using this to show the project title, etc. Is there a way to show it but without the dropdown?

orchestra/static/orchestra/project-management/data-service.es6.js
```diff
@@ -43,15 +42,13 @@ export default function dataService ($location, $rootScope, $route, orchestraApi
...
-        this.currentProject = this.allProjects[projectId]
-        return this.setSelectedProject()
-      }.bind(this))
+      if (!projectId) {
+        $route.updateParams({projectId: undefined})
+        return
+      }
+      this.currentProject.id = projectId
+      this.resetData()
+      return this.updateData()
```
marcua, line 50: curiously, why remove `this.ready` to make sure we have the data?

orchestra/static/orchestra/project-management/data-service.es6.js
```diff
@@ -82,6 +79,7 @@ export default function dataService ($location, $rootScope, $route, orchestraApi
       /**
        * Prepares raw project data for visualization.
        */
+      this.currentProject = data[this.currentProject.id].project
```
marcua, line 81: what's this change for?

orchestra/static/orchestra/project-management/data-service.es6.js
```diff
@@ -29,8 +29,7 @@ export default function dataService ($location, $rootScope, $route, orchestraApi
     },
     setSelectedProject: function () {
       $route.updateParams({projectId: this.currentProject.id})
-      this.resetData()
-      return this.updateData()
+      return this.changeProject(this.currentProject.id)
```
marcua, line 31: what's the refactor for?

orchestra/static/orchestra/project-management/data-service.es6.js
adbharadwaj: We were earlier loading all projects in `this.ready`. Actually, we were repurposing ready to either load all projects or load Project info. It was very confusing. I have got rid of all usages of ready and call specific functions instead.

orchestra/static/orchestra/project-management/data-service.es6.js
adbharadwaj: I just found it weird we were calling `updateData` in  `setSelectedProject`. Since, users will not be able to use the dropdown anymore when a project is selected. I felt `updateData` should only be called in changeProject.

orchestra/static/orchestra/project-management/data-service.es6.js
adbharadwaj: Added comment to explain!

orchestra/static/orchestra/project-management/project-management.html
adbharadwaj: Fixed!

orchestra/static/orchestra/project-management/project-vis.es6.js
adbharadwaj: Got rid of all usage of the ready function to add more clarity! We don't have to return anything!

marcua (approved):
Nice job on cleaning up and explaining the changes!

Approving, but in your test videos I didn't see you interacting with the buttons on a project's page (admin, add to Slack channel, the revert flags, etc.). There's no reason to believe they are broken, and you don't need a video of it, but just click around and make sure the interface itself still works :).

PR #808, opened 2021-09-27
Add required field to todo templates

jumasheff:


orchestra/todos/import_export.py
```diff
@@ -64,7 +66,7 @@ def export_to_spreadsheet(todo_list_template):
     """
```
marcua, line 66: worth adding a required to-do and
* ensuring that all to-dos other than it are not required
* it is required?

orchestra/todos/import_export.py
```diff
@@ -16,6 +16,7 @@
 REMOVE_IF_HEADER = 'Remove if'
```
marcua, line 16: In unit tests, it's worth adding a required to-do and ensuring that
* all to-dos other than it are not required
* it is required

orchestra/todos/import_export.py
```diff
@@ -101,10 +103,11 @@ def import_from_spreadsheet(todo_list_template, spreadsheet_url, request):
             'remove_if': json.loads(row[1] or '[]'),
             'skip_if': json.loads(row[2] or '[]'),
             'slug': row[0] or None,
+            'required': row[3] and row[3].lower() == 'true' or False,
```
marcua: Curiously, what's the `or False` for? I might be silly, but in my head `X or False == X`?

orchestra/todos/api.py
```diff
@@ -100,7 +100,8 @@ def _add_template_todo(
             template=todolist_template,
             parent_todo=parent_todo,
             status=status,
-            additional_data=additional_data
+            additional_data=additional_data,
+            required=template_todo.get('required', False)
```
marcua, line 104: Don't overthink this---just starting a discussion that I don't have a strong opinion about: are we overloading concepts too much? We have to-do slugs and we have `required`, and in some cases (internal codebase) we rely on the slug to identify its qualities and others we rely on a specific property like `required`.
* Argument for implementing this PR: `required` is a pretty foundational concept, and stands alone.
* Argument against implementing this PR: We already rely on the slug to identify required status in some internal functionality like action links---do we want to have two ways to make a to-do required?

jumasheff: @marcua My thoughts about `required` and `slug`:

General cases: in certain cases we want to treat todos as required except for a small number of todos

- pros: we already have some logic which checks if required todos are completed, as in, the `required` field is already being used and we can happily reuse it

Special cases: in some cases we want to treat a smaller set of todos as required

- pros: since all todos except some edge cases are marked as required, the `required` field is not helpful here. Here we want to use the `slug`

orchestra/todos/import_export.py
jumasheff: `or False` it covers a case when there's an empty cell in `row[3]`. That is, if we omit `or False`, it:

- sets True (pythonic true) when it encounters `TRUE` (spreadsheet converts True/true into TRUE) 
- it sets just an empty string when it encounter an empty string

But in case we add `or False`, it sets False to `required` instead of an empty string.

jumasheff: Please let me know if treating empty strings as False is a bad idea.

marcua: That was helpful! Thank you! Given it confused me when I read it, I wonder if a more explicit version with two boolean values would be clearer: `isinstance(row[3], str) and row[3].lower() == 'true'` evaluates two booleans, and doesn't require the `or False`. What do you think?

orchestra/todos/api.py
adbharadwaj: My argument for adding it was the same as @marcua listed -- `required is a pretty foundational concept, and stands alone.`

adbharadwaj:
code lgtm!

marcua:
Thank you for iterating, lgtm! :)

PR #810, opened 2021-09-30
Bug fix: Root template todo to use slug, required value given in the template

adbharadwaj:
- Fixes a bug where root template todo was not using the slug, required value given in the template


jumasheff:
A! You can pull updates from `main` to make build stage pass.

PR #814, opened 2021-10-06
Mark the whole todo tree as deleted

jumasheff:


orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -93,9 +105,10 @@ export default function todoList (orchestraApi) {
       }
 
       todoList.removeTodo = (todo) => {
```
marcua, line 107: for consistency, would this behavior also apply to declining/undeclining nested to-dos?

orchestra/static/orchestra/todos/todo-list.directive.es6.js
```diff
@@ -93,9 +105,10 @@ export default function todoList (orchestraApi) {
...
-        var index = todoList.todos.indexOf(todo)
-        todoList.todos.splice(index, 1)
-        todoApi.delete(todo)
+        const idsToUpdate = getNestedTodoIds(todo)
+        remove(todoList.todos, t => idsToUpdate.includes(t.id))
```
marcua, line 109: if `N = todoList.todos.length` and `M = idsToUpdate.length`, this is an O(N*M) search, whereas if `idsToUpdate` was a set instead of a list, it would be O(N)

adbharadwaj: +1

PR #816, opened 2021-12-09
Update tasktable to show next steps in paused tasks

elstonayx:


orchestra/__init__.py
```diff
@@ -1,4 +1,4 @@
 # The current Orchestra version.
-__version__ = '1.0.49'
+__version__ = '1.0.50'
```
marcua: Thank you for being proactive :). We don't bump versions this way --- we have a script (https://github.com/b12io/orchestra/blob/main/release.py) that will handle this once the code is on `main`.

orchestra/utils/task_lifecycle.py
```diff
@@ -549,7 +549,7 @@ def tasks_assigned_to_worker(worker):
             next_todo_title = None
             next_todo_dict = {}
             should_be_active = False
-            if state in ('returned', 'in_progress'):
+            if state in ('returned', 'in_progress', 'paused'):
```
marcua, line 552: Wow that was simple! :) If you rested this in dev with production data, and didn't run into issues, that's great! :)

marcua: One thing to watch out for: I assume it's OK, but do we only use `should_be_active` (set below) for `returned`/`in_progress` tasks, or do we have to be careful about how it might be used in the `paused` table?

elstonayx: I'll add a guard to check if it's a paused task so that we can have the same behavior as before for `should_be_active`.

marcua: LGTM assuming you compared on real data before/after! Thanks for programming defensively! :)

ashleyhlai (approved):

PR #830, opened 2022-02-15
Some tests misusing assertTrue for comparisons fix

### LINKED ISSUE #829: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments, should use `assertEqual` for that.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is assertEqual. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/b12io/orchestra/blob/c407f7372a09f086ef168d5994b186e83731fa8e/orchestra/tests/test_dashboard.py#L63
https://github.com/b12io/orchestra/blob/c407f7372a09f086ef168d5994b186e83731fa8e/orchestra/tests/test_query_params_filter.py#L90

I found this issue automatically, see other issues [here](https://codereview.doctor/b12io/orchestra)

code-review-doctor:
Fixes #829

### DIFF
--- a/orchestra/tests/test_dashboard.py
+++ b/orchestra/tests/test_dashboard.py
@@ -60,7 +60,7 @@ class DashboardTestCase(OrchestraTransactionTestCase):
         self.assertTrue(machine_task_assignment.status,
                         TaskAssignment.Status.SUBMITTED)
 
-        self.assertTrue(machine_task_assignment.in_progress_task_data,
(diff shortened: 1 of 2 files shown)


marcua:
Thank you doctor ;)

PR #839, opened 2022-05-19
Fix "Unsupported media type \"\" in request." in `delete_todos`

jumasheff:


orchestra/orchestra_api.py
```diff
@@ -128,7 +128,9 @@ def update_todos(updated_todos):
 
 
 def delete_todos(todo_ids):
+    headers = {'Content-Type': 'application/json; charset=UTF-8'}
     response = _make_api_request('delete', 'todo-api',
```
adbharadwaj, line 132: Can it be same as what we have in update_todos? (just to be consistent?)

PR #855, opened 2023-01-13
Show task submit button only to project admins

adbharadwaj:
This PR ensures that task submit button is only visible to project admins now.


orchestra/static/orchestra/task/task.html
```diff
@@ -114,7 +114,8 @@ <h3 class="modal-title">{{command | capitalize}} Confirmation</h3>
     <div class="container-fluid">
```
marcua, line 114: This is great! Just so we don't confuse new users, we'll want to make sure we create the group and add it to the sample workflow data:
* [Our docs](https://orchestra.readthedocs.io/en/latest/example_use.html#setup) list the accounts that are created
* [This script](https://github.com/b12io/orchestra/blob/main/journalism_workflow/load_sample_data.py) does the loading
* I think the fix is to `from django.contrib.auth.models import Group` at the top of the script and add something like
```
project_admins = Group.objects.get_or_create(name='project_admins') 
project_admins.user_set.add(user)
```
after [this line](https://github.com/b12io/orchestra/blob/db95c574cf8c3aab42f63932fc6ceb995c10634a/journalism_workflow/load_sample_data.py#L21)

marcua (approved):
lgtm, assuming you tested the two lines (even at the console is OK for our purposes)

simple_workflow/load_sample_data.py
```diff
@@ -21,5 +21,7 @@ def load(workflow_version):
     user.save()
     if created:
         Worker.objects.create(user=user)
-    project_admins = Group.objects.get_or_create(name='project_admins')
-    project_admins.user_set.add(user)
+    project_admins, created = Group.objects.get_or_create(
+        name='project_admins')
+    if created:
```
marcua: even if the group already existed, we should add the user to the group

journalism_workflow/load_sample_data.py
```diff
@@ -20,8 +20,10 @@ def load(workflow_version):
         user.set_password(password)
         user.save()
         worker, _ = Worker.objects.update_or_create(user=user)
-        project_admins = Group.objects.get_or_create(name='project_admins')
-        project_admins.user_set.add(user)
+        project_admins, created = Group.objects.get_or_create(
+            name='project_admins')
+        if created:
```
marcua: same comment

marcua (changes requested):

marcua (approved):

PR #866, opened 2025-10-30
Django 5.2

cadaxapy:


example_project/example_project/orchestra_settings.py
```diff
@@ -117,9 +117,7 @@ def setup_orchestra(settings_module_name):
     # Hijack settings
     settings.INSTALLED_APPS += (
         'hijack',
-        'compat',
-        'hijack_admin',
-        'related_admin',
+        'hijack.contrib.admin'
```
thisisdhaas, line 120: did you test this?

example_project/example_project/urls.py
thisisdhaas, line 1: Similar to the product PR, I'm skipping urls.py files, lmk if that's not the right thing to do.

orchestra/static/orchestra/common/css/_common.scss
```diff
@@ -76,6 +76,92 @@ $main-font-stack: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helveti
 }
 
 // styles related to the hijack banner
+.hijacked-warning {
```
thisisdhaas, line 79: where did these styles come from?

thisisdhaas: (and why didn't we need something similar in the product PR?)

thisisdhaas: oh wait nm it is in the product PR. I won't leave a comment there, so you can let me know here what it's all about :-)

cadaxapy: In the latest versions of django-hijack, the notification-bar styles were removed, so I had to manually copy them here.

example_project/example_project/orchestra_settings.py
cadaxapy: Yes, there’s just one visual change: in the admin, it now says 'Impersonate' instead of 'Hijack'.

example_project/example_project/urls.py
cadaxapy: The only changes in urls.py are the url → re_path renames, so it can be skipped.

orchestra/static/orchestra/common/css/_common.scss
thisisdhaas: got it, ty!

thisisdhaas (approved):

